Issue #20, opened 2023-10-02
Create address to transfer component


wainola:
## Implementation details

- by default it transfer to the same address that the wallet is connected, if destination chain is Evm based
- if destination chain is not Evm, toggle should not be rendered
- if toggle is active, input to add the address should be render
- we should apply styles defined in figma

## Acceptance Criteria

- if toggle is active, input must be displayed

wainola:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Sygma-Protocol-62f126a54b7d5a55203732f0/issues/sygmaprotocol/sygma-widget/20?planning-poker) with Zenhub @FSM1 @wainola

Lykhoyda:
@wainola is there a link for a design/figma?

PR #56, opened 2024-01-18
feat: address to transfer

wainola:
## Description

This PR adds toggle and input to introduce address to transfer

## Related Issue Or Context

Closes: #20 

## Types of changes

- [ ] Documentation

## Checklist:


packages/widget/src/components/address-input/address-input.ts
```diff
@@ -0,0 +1,32 @@
...
+import { ifDefined } from 'lit/directives/if-defined.js';
+import { styles } from './styles';
+
+@customElement('address-input')
+export default class AddressInput extends LitElement {
+  static styles = styles;
+  @property({
+    type: String
+  })
+  addressToTransfer?: string;
```
mpetrunic, line 13:
```suggestion
  address?: string;
```

packages/widget/src/components/address-input/address-input.ts
```diff
@@ -0,0 +1,32 @@
...
+
+  @property()
+  handleAddress?: (e: Event) => void;
+
+  render(): HTMLTemplateResult {
+    return html`<section class="switch-container">
```
mpetrunic, line 19:
```suggestion
    return html`<section class="address-input-container">
```

packages/widget/src/components/address-input/address-input.ts
```diff
@@ -0,0 +1,32 @@
+import { LitElement, html } from 'lit';
+import type { HTMLTemplateResult } from 'lit';
+import { customElement, property } from 'lit/decorators.js';
+import { ifDefined } from 'lit/directives/if-defined.js';
+import { styles } from './styles';
+
+@customElement('address-input')
+export default class AddressInput extends LitElement {
```
mpetrunic, line 8: I think the component should handle address validation

packages/widget/src/components/address-input/address-input.ts
```diff
@@ -0,0 +1,32 @@
+import { LitElement, html } from 'lit';
+import type { HTMLTemplateResult } from 'lit';
+import { customElement, property } from 'lit/decorators.js';
+import { ifDefined } from 'lit/directives/if-defined.js';
+import { styles } from './styles';
+
+@customElement('address-input')
```
mpetrunic, line 7: considering all elements live in shared namespace, I would suggest prefixing with "sygma-"

packages/widget/src/components/address-input/address-input.ts
```diff
@@ -0,0 +1,32 @@
...
+export default class AddressInput extends LitElement {
+  static styles = styles;
+  @property({
+    type: String
+  })
+  addressToTransfer?: string;
+
+  @property()
+  handleAddress?: (e: Event) => void;
```
mpetrunic, line 16:
```suggestion
  handleAddress?: (address: string) => void;
```
make sure component parses and validates before passing address to callback

packages/widget/src/components/address-input/address-input.ts
```diff
@@ -0,0 +1,32 @@
...
+
+  render(): HTMLTemplateResult {
+    return html`<section class="switch-container">
+      <div class="input-address-container">
+        <label>Send to</label>
+        <input
+          class="input-address"
+          name="address"
+          type="text"
+          @change=${this.handleAddress}
```
mpetrunic, line 26: create and call private method that validates input and invokes handleAddress callback

packages/widget/src/components/address-input/index.ts
```diff
@@ -0,0 +1 @@
+export { default } from './address-input';
```
mpetrunic, line 1: can we use named exports, those default exports have really weak support in IDEs

mpetrunic (changes requested):

packages/widget/src/components/address-input/address-input.ts
wainola: got it

mpetrunic:
it might make sense to rebase against: https://github.com/sygmaprotocol/sygma-widget/pull/62 so you can add address input tests.

wainola:
> it might make sense to rebase against: #62 so you can add address input tests.

sure, I'm going to do it

packages/widget/src/components/address-input/address-input.ts
```diff
@@ -0,0 +1,63 @@
...
+  })
+  address: string = '';
+
+  @property()
+  handleAddress?: (address: string) => void;
+
+  @property({
+    type: Object
+  })
+  network?: 'substrate' | 'evm';
```
wainola, line 23: @mpetrunic @sztok7 here I would like to use `Network` enum from `sygma-sdk` but somehow changing this, and then passing network property as `${Network.EVM}` on the test, gives error. I'm going to check that tomorrow. Question here: should the input be aware of the network to branch on validation?

packages/widget/src/components/address-input/address-input.ts
```diff
@@ -0,0 +1,63 @@
...
+
+  private handleAddressChange = ({ target }: Event): void => {
+    const { value } = target as HTMLInputElement;
+    if (this.network !== 'evm') {
+      const validPolkadotAddress = validatePolkadotAddress(value);
+
+      if (validPolkadotAddress) {
+        return void this.handleAddress?.(value);
+      }
+      // TODO: if not defined or show error
```
wainola, line 33: @mpetrunic do we have any definition for error messages?

packages/widget/tests/unit/components/address-input/address-input.test.ts
```diff
@@ -0,0 +1,87 @@
+import { afterEach, assert, describe, it, vi } from 'vitest';
+import { fixture, fixtureCleanup, oneEvent } from '@open-wc/testing-helpers';
+import { html } from 'lit';
+import { AddressInput } from '../../../../src/components';
+
+describe('address-input component', function () {
```
mpetrunic, line 6: add test cases for invalid ethereum/sustrate address

mpetrunic (commented):
looking great so far, lets add validation :)

packages/widget/src/components/address-input/address-input.ts
wainola: I changed to `string` to test implementation. This is the error that gives me when using the enum:

```
Error: Cannot find module '/Users/wainola/Documents/js/chainsafe/sygma-widget/node_modules/@buildwithsygma/sygma-sdk-core/dist-esm/localConfig' imported from /Users/wainola/Documents/js/chainsafe/sygma-widget/node_modules/@buildwithsygma/sygma-sdk-core/dist-esm/config.js
```

Which is weird since `localConfig` or event `config` don't have anything to do with the importing of the enum from the SDK

packages/widget/src/utils/index.ts
```diff
@@ -46,3 +48,12 @@ export const capitalize = (s: string): string => {
   const rest = s.slice(1);
   return `${firstLetter}${rest}`;
 };
+
+export const validatePolkadotAddress = (address: string): boolean => {
```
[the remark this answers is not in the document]
wainola: btw @mpetrunic because we hare using `jsdom` as env for the test, the call to `decodeAddress(address)` function fails because of `Invalid decoded address checksum`. I tested changing the env to node and it passes. Also if you pass `false` to avoid checking checksum it also passes. Here I'm hesitant on how to proceed, do I mock `decodeAddress` just to make the tests pass? Running this on browser context works as expected, since this is the recipe from the `polkadot.js` docs

packages/widget/src/components/address-input/address-input.ts
mpetrunic: ah, thats bug on sygma-sdk. There is a lot of places which are missing .js extension in import statements.
Issue: https://github.com/sygmaprotocol/sygma-sdk/issues/326 I'll fix that today^^

wainola: yeah that's the only diff with the other imports from `config` file

packages/widget/src/components/address-input/address-input.ts
wainola: Implemented and shared with Cindy

packages/widget/src/components/address-input/address-input.ts
wainola: new PR adding small fixing that was appearing on test [pr](https://github.com/sygmaprotocol/sygma-sdk/pull/329)

packages/widget/src/utils/index.ts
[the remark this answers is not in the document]
mpetrunic: @wainola pushed commit to change happy-dom to jsdom (it seems to not have this issue) but we should work on running in actual browser.

mpetrunic:
@wainola Can you put this component inside widget?

wainola:
> @wainola Can you put this component inside widget?

it is, I haven't pushed that. Will do shortly

packages/widget/src/components/address-input/address-input.ts
```diff
@@ -0,0 +1,87 @@
...
+@customElement('sygma-address-input')
+export class AddressInput extends LitElement {
+  static styles = styles;
+  @property({
+    type: String
+  })
+  address: string = '';
+
+  @property({ attribute: false })
+  onHandleAddress?: (address: string) => void;
```
sztok7, line 20: maybe `onAddressChange` would make a bit more sense? also, not sure if this prop should be optional?

packages/widget/src/components/address-input/address-input.ts
```diff
@@ -0,0 +1,87 @@
...
+        this.errorMessage = 'Invalid Substrate address';
+        return;
+      }
+    } else {
+      const isAddress = ethers.utils.isAddress(value);
+      if (!isAddress) {
+        this.errorMessage = 'Invalid Ethereum Address';
+        return;
+      }
+    }
```
sztok7, line 59: not a big fan of ifs with negative logic, because they obfuscate code. i'd rather put `this.network === Network.SUBSTRATE` as condition or just `this.network === Network.EVM` and switch clauses content

packages/widget/src/components/address-input/address-input.ts
```diff
@@ -0,0 +1,87 @@
...
+    return void this.onHandleAddress?.(value);
```
sztok7, line 60: is this `return` necessary?

packages/widget/src/components/address-input/styles.ts
```diff
@@ -0,0 +1,36 @@
...
+    flex-direction: column;
+    justify-content: center;
+    gap: 8px;
+  }
+
+  .inputAddressContainer {
+    display: flex;
+    flex-direction: column;
+    width: 100%;
+    height: 91px;
```
sztok7, line 15: not a big fan of 'hardcoded' values for height/width as they make style harder to maintain. we can discuss other approaches on slack :)

packages/widget/src/components/address-input/styles.ts
```diff
@@ -0,0 +1,36 @@
...
+    gap: 8px;
+  }
+
+  .inputAddress {
+    border-radius: 24px;
+    border: 1px solid var(--zinc-200, #e4e4e7);
```
sztok7, line 21: copy pasta from figma? 😂 

```suggestion
    border: 1px solid #e4e4e7;
```

also a suggestion: add `variables.ts` file with color variables and use them in files (follow Cindy's convention of naming them, e.g. `const zinc400 = '#e4e4e7'`)

sztok7 (changes requested):

packages/widget/src/components/address-input/address-input.ts
wainola: shouldn't actually

packages/widget/src/components/address-input/styles.ts
wainola: I think we should create an issue addressing styling overall over the widget, so we do not loose the scope of this PR. 100% agree

packages/widget/src/components/address-input/address-input.ts
wainola: not really

mpetrunic:
@wainola since we still don't have designs for error states, could you just:
- make error message font smaller and less bold
- make sure it doesn't move components around when displayed
- ensure it aligns with right side
- maybe make message lowercase?

Also I think input border is too accented when focused and input text font is bigger in the design

packages/widget/src/components/address-input/styles.ts
wainola: applied color definition to `:host` selector

wainola:
> @wainola since we still don't have designs for error states, could you just:
> 
> * make error message font smaller and less bold
> * make sure it doesn't move components around when displayed
> * ensure it aligns with right side
> * maybe make message lowercase?
> 
> Also I think input border is too accented when focused and input text font is bigger in the design

Fixed, although we still should have a theming/style issue once we have everything there, to check changes on size when widget is connected

packages/widget/src/components/address-input/address-input.ts
sztok7: and what about changing name to something a bit more semantic?

packages/widget/src/components/address-input/address-input.ts
```diff
@@ -0,0 +1,90 @@
...
+    type: String
+  })
+  network: Network = Network.EVM;
+
+  @state()
+  errorMessage?: string;
+
+  connectedCallback(): void {
+    super.connectedCallback();
+    if (this.address) {
```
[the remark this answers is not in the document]
mpetrunic: Are you sure handleAddressChange triggers if the address is prefilled?

packages/widget/src/components/address-input/address-input.ts
```diff
@@ -0,0 +1,90 @@
...
+        <label>Send to</label>
+        ${when(
+          this.errorMessage,
+          () => html` <span class="errorMessage">${this.errorMessage}</span>`
+        )}
+        <input
+          class=${this.errorMessage ? 'inputAddress error' : 'inputAddress'}
+          name="address"
+          type="text"
+          @change=${(evt: Event) => this.handleAddressChange.bind(this)(evt)}
```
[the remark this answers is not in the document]
wainola: actually yes and for some reason I left this unchanged

packages/widget/src/components/address-input/address-input.ts
```diff
@@ -0,0 +1,90 @@
...
+  }
+
+  private handleAddressChange = ({ target }: Event): void => {
+    const { value } = target as HTMLInputElement;
+
+    if (this.errorMessage) {
+      this.errorMessage = undefined;
+    }
+
+    if (value !== '') {
```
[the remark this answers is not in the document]
wainola: I agree, but the way I see it is that we want to display the error message either if the value is empty string or an invalid/valid address. Since this is not going to be undefined, I can't imagine how to do an early return over this. We don't want this to pass for empty string to the callback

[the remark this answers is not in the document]
wainola: @sztok7 let me know if the pushed solution makes sense to you

packages/widget/src/components/address-input/address-input.ts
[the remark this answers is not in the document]
wainola: @mpetrunic yes it does, why you ask?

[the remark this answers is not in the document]
wainola: @sztok7 makes sense what you mention here

packages/widget/src/components/address-input/address-input.ts
```diff
@@ -0,0 +1,90 @@
...
+        }
+      }
+    }
+
+    void this.onAddressChange?.(value);
+  };
+
+  render(): HTMLTemplateResult {
+    return html`<section class="addressInputContainer">
+      <div class="inputAddressContainer">
```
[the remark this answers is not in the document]
wainola: no, we haven't. Going to change that.

packages/widget/src/styles.ts
```diff
@@ -15,11 +26,11 @@ export const styles = css`
     gap: 16px;
 
     padding: 24px;
```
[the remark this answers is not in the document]
wainola: yeah indeed, going to check either if there is an issue to address that or create one

packages/widget/src/components/address-input/address-input.ts
```diff
@@ -0,0 +1,90 @@
...
+          this.errorMessage = 'invalid Substrate address';
+          return;
+        }
+      } else {
+        const isAddress = ethers.utils.isAddress(value);
+        if (!isAddress) {
+          this.errorMessage = 'invalid Ethereum Address';
+          return;
+        }
+      }
```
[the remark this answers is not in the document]
wainola: done

packages/widget/src/components/address-input/address-input.ts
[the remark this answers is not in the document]
wainola: btw @mpetrunic I think we should also validate that the input is not empty upon submission. If not mistaken the flow is going to be: wallet connected => input prefilled with connected wallet address. But if the user changes that and the input remains empty, I think we should warn the user + don't allow him to submit. Wdyt?

packages/widget/src/components/address-input/address-input.ts
```diff
@@ -0,0 +1,90 @@
...
+
+  @state()
+  errorMessage?: string;
+
+  connectedCallback(): void {
+    super.connectedCallback();
+    if (this.address) {
+      this.handleAddressChange({
+        target: { value: this.address }
+      } as unknown as Event);
```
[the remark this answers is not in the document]
wainola: for this, it means we would need to do this:

```ts
 @change=${(evt: Event) =>
            this.handleAddressChange((evt.target as HTMLInputElement).value)}
```

I don't think this is hacky, is just type gymnastics sponsored by TS. Since we use this function to also validate the prefilled address upon first render of the component. Either way I think both works. What are your thoughts?

packages/widget/src/components/address-input/address-input.ts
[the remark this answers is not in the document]
mpetrunic: We should, but out of scope for this PR. Widget will handle that (disabling submit, writing some error message)

mpetrunic:
@wainola One small issue, when error message shows it pushes input field down. Could you ensure this doesn't happen (align message on the right side of the input field)

packages/widget/src/components/address-input/address-input.ts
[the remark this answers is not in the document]
sztok7: > I don't think this is hacky, is just type gymnastics sponsored by TS.

hahahahahahaha isn't `type gimnastics` literally definition of being hacky 😆 

i'd rather unwrap event like you wrote and just take string as argument in that method, since obviously it can be called without emitting an Event

packages/widget/src/components/address-input/styles.ts
```diff
@@ -0,0 +1,45 @@
...
+    min-height: 7.75rem;
+    height: 100%;
+    gap: 0.5rem;
+  }
+
+  .inputAddress {
+    border-radius: 1.5rem;
+    border: 0.063rem solid var(--zinc-200);
+    width: 100%;
+    height: 4.25rem;
```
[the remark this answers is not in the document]
sztok7: i find padding more future-proof. if everything is setup properly, height will be that same as if you used fixed height and you don't have to worry about changing height if let's say font-size on input changes. i always opt for 'fluid' options, because fixed ones usually have some edge-cases that break them

packages/widget/src/components/address-input/address-input.ts
```diff
@@ -0,0 +1,84 @@
...
+    return html` <section class="inputAddressSection">
+      <div class="inputAddressContainer">
+        <label class="labelContainer">
+          <span>Send to </span>
+          ${when(
+            this.errorMessage,
+            () => html` <span class="errorMessage">${this.errorMessage}</span>`
+          )}</label
+        >
+        <textarea
```
wainola, line 64: @mpetrunic yesterday after some discussion we arrive to one posible solution to the problem of not being able to make word-break for susbtrate address: tu use textarea. Minor changes here is that we need to trim the value before sending it to validation, since values of text area have a next line character

packages/widget/src/components/address-input/address-input.ts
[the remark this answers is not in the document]
wainola: changed

packages/widget/src/components/address-input/address-input.ts
mpetrunic: Im fine with textarea

packages/widget/src/components/address-input/address-input.ts
```diff
@@ -0,0 +1,84 @@
...
+          class=${this.errorMessage ? 'inputAddress error' : 'inputAddress'}
+          name="address"
+          @change=${(evt: Event) =>
+            this.handleAddressChange(
+              (evt.target as HTMLInputElement).value.trim()
+            )}
+          rows="2"
+        >
+${ifDefined(this.address)}</textarea
```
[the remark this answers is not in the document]
mpetrunic: no idea, probably prettier is confused with syntax and doesn't lint it

mpetrunic (approved):

mpetrunic (changes requested):
powered by dropped out of widget and add trim to the input^^

packages/widget/src/components/address-input/address-input.ts
Lykhoyda: The text area looks strange, especially since we can put the cursor in any place:

Screenshot 2024-01-31 at 17 27 53

I didn't get the problem with using the input apart that the entire address value will not fit into the field. But from my experience, it's better than showing the first and last 4 characters f.e:
5E75.....8cem
Did I miss something?

packages/widget/src/components/address-input/address-input.ts
```diff
@@ -0,0 +1,83 @@
...
+      void this.onAddressChange(trimedValue);
+    }
+  };
+
+  render(): HTMLTemplateResult {
+    return html` <section class="inputAddressSection">
+      <div class="inputAddressContainer">
+        <label class="labelContainer">
+          <span>Send to </span>
+          ${when(
```
Lykhoyda, line 59: As per my understanding each render "when" will be treated as a new function and will lead to unnecessary re-render. So we might define it in the class f.e:

```
private renderErrorMessage() {
 return when()
}
```

packages/widget/src/components/address-input/address-input.ts
```diff
@@ -0,0 +1,83 @@
...
+          )}</label
+        >
+        <textarea
+          class=${this.errorMessage ? 'inputAddress error' : 'inputAddress'}
+          name="address"
+          @change=${(evt: Event) =>
+            this.handleAddressChange((evt.target as HTMLInputElement).value)}
+          rows="2"
+        >
+${ifDefined(this.address)}
```
Lykhoyda, line 71: indeed prettier is not formatting properly, but if we manually move the value to the middle with spaces it will not complain 😅

packages/widget/src/components/address-input/address-input.ts
wainola: interesting, do you have a source for this? I wasn't aware

packages/widget/src/components/address-input/address-input.ts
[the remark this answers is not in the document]
sztok7: @wainola shall we remove this redundant check then?

packages/widget/src/components/address-input/styles.ts
```diff
@@ -0,0 +1,51 @@
+import { css } from 'lit';
+
+export const styles = css`
+  .inputAddressSection {
+    display: flex;
+    flex-direction: column;
+    justify-content: center;
+    gap: 0.5rem;
+    min-height: 7.75rem;
```
sztok7, line 9: these `min-height` props (here and next class) are used only for spacing between elements which is pretty wrong. i would remove them and set appropriate gap/margin.

no need to change this now (so PR can be merged), but please then add `TODO:`, so I/we can update CSS in a day or two.

sztok7: we can do that together in a call, so I can explain a bit further the approach with no setting fixed values fro height/width

packages/widget/src/components/address-input/address-input.ts
```diff
@@ -0,0 +1,83 @@
...
+            this.errorMessage,
+            () => html` <span class="errorMessage">${this.errorMessage}</span>`
+          )}</label
+        >
+        <textarea
+          class=${this.errorMessage ? 'inputAddress error' : 'inputAddress'}
+          name="address"
+          @change=${(evt: Event) =>
+            this.handleAddressChange((evt.target as HTMLInputElement).value)}
+          rows="2"
```
sztok7, line 69: default is 2, so this is redundant, also, this just sets textarea height to be 2 rows high (which is overridden by setting height in px)

packages/widget/src/components/address-input/address-input.ts
```diff
@@ -0,0 +1,83 @@
...
+  @property({ attribute: false })
+  onAddressChange: (address: string) => void = () => {};
+
+  @property({
+    type: String
+  })
+  network: Network = Network.EVM;
+
+  @state()
+  errorMessage?: string;
```
sztok7, line 27: i suggest making this non-optional (since we don't need a three-state, undefined/''/'sum ting wong') and if it's an empty string (falsy value), all good, if no, error!

packages/widget/src/components/address-input/address-input.ts
```diff
@@ -0,0 +1,83 @@
...
+    if (this.address) {
+      this.handleAddressChange(this.address);
+    }
+  }
+
+  private handleAddressChange = (value: string): void => {
+    const trimedValue = value.trim();
+
+    if (this.errorMessage) {
+      this.errorMessage = undefined;
```
sztok7, line 40: here we'd need to change this to
```suggestion
      this.errorMessage = null;
```

packages/widget/src/components/address-input/address-input.ts
sztok7: even better, to be more explicit, i think it'd be better to put 
```suggestion
  errorMessage: string | null = null;
```

packages/widget/src/utils/index.ts
```diff
@@ -46,3 +50,39 @@ export const capitalize = (s: string): string => {
...
+      }
+      break;
+    }
+    default:
+      errorMessage = 'invalid address';
+      break;
+  }
+
+  return errorMessage;
+};
```
sztok7, line 88: and lastly, this function would look something like this:
```suggestion
export const validateAddress = (address: string, network: Network): string | null => {
  switch (network) {
    case Network.SUBSTRATE: {
      const validPolkadotAddress = validateSubstrateAddress(address);

      return validPolkadotAddress ? null : 'invalid Substrate address';
    }
    case Network.EVM: {
      const isAddress = ethers.utils.isAddress(address);
      
      return isAddress ? null : 'invalid Ethereum address';
    }
    default:
      return 'unsupported network';
  }
};
```

sztok7 (changes requested):

packages/widget/src/components/address-input/styles.ts
wainola: adding gap on the form using flex could be a solution

Lykhoyda (approved):

sztok7 (approved):

mpetrunic (approved):

Issue #10, opened 2023-09-13
Create React package


wainola:
Lit allows us to wrap our web component into a react one. Hence, in a different package we can define a react component to be used for react frontends, for ease of use.

## Implementation details

- create package that imports `Lit` widget and wraps it into a react component
- provide unit test of the react component working properly
- create "examples" directory in the root of the repo and add it to the workspaces
- in examples directory create simple example nextjs app and install sygma react widget inside

## Testing details

- imported react component is able to render on basic react app

## Acceptance Criteria

- widget example using react component

wainola:
Current progress of this task: definition of the interface of the widget that is going to be exposed into the react package

PR #61, opened 2024-01-22
feat: base interface implementation for react package

wainola:
## Related Issue Or Context

Closes: #10 

## Types of changes

- [ ] Documentation

## Checklist:


packages/widget/src/widget.ts
```diff
@@ -1,14 +1,37 @@
...
+
+  @property({
+    type: Object
+  })
+  web3Provider?: ethers.providers.Web3Provider;
+
+  @property({
+    type: Object
+  })
+  apiPromise?: ApiPromise;
```
mpetrunic, line 28: maybe rename to "substrateProvider"? I would also merge ApiPromise and wssConnectionUrl into single property

packages/widget/src/widget.ts
```diff
@@ -1,14 +1,37 @@
...
+import { customElement, property } from 'lit/decorators.js';
 import './components/widget-app';
+import type { ethers } from 'ethers';
+import type { ApiPromise } from '@polkadot/api';
 import { styles } from './styles';
 import { Network } from './controllers';
+import type { ISygmaProtocolWidget } from './interfaces';
 
 @customElement('sygmaprotocol-widget')
 class SygmaProtocolWidget extends LitElement {
```
mpetrunic, line 12: I think we also need:
- "show" od "disable" attribute.
- "whitelistedResource" property to enable dapp to lock widget to only transfer their token. Maybe in https://chainagnostic.org/CAIPs/caip-19

packages/widget/src/widget.ts
```diff
@@ -1,14 +1,37 @@
...
   static styles = styles;
 
+  @property({
+    type: Object
+  })
+  network?: Network;
```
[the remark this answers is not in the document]
MakMuftic: I agree here, having a whitelist for source and destination chains is a functionality we want to support :+1:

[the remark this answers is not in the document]
MakMuftic: In regards to the CAIP suggestion @mpetrunic, supporting this makes a lot of sense (I would love it if this was on our radar before). 
My concern here is if this makes more sense on the SDK level, as we will always need to remap it back to Sygmas resourceIDs and domainIDs.

packages/widget/src/widget.ts
wainola: If I understand you correctly, susbtrateProvider could be either ApiPromise or string?

packages/widget/src/widget.ts
wainola: Would it make sense to also define at the interface level, style options for the widget to pass as properties? We discussed about this with @MakMuftic but we didin't arrive to any agreement and I guess it makes sense, since we are defining the interfaces here.

mpetrunic: yes!

packages/widget/src/widget.ts
mpetrunic: the goal is to enable dapp to provider substrate related stuff as object in single property. So dapp should be able to, optionally, provide signer and provider.
So the type could be:
```typescript
interface {
  provider?: ApiPromise|string,
  signer: NotSure //it needs to support external signer like extension
}
˙``

File: packages/widget/src/widget.ts (hunk 3 above)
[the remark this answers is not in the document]
Comment (mpetrunic) [reply packages/widget/src/widget.ts:18]: It should be possible to map caip to sygma resource id in SDK, no?
I'm thinking if we wanna do that change now, else we will have to cut major version for widget soon.

File: packages/widget/src/widget.ts (hunk 1 above)
Comment (mpetrunic) [reply packages/widget/src/widget.ts:28]: thibaut says:
`˙`
api: ApiPromise | string
signer: Signer // that's actually a type from '@polkadot/api/types'
senderAddress: string // the signer doesn't have the address easy to get afaics
```

wainola: how the user flow is going to work passing signer or sender address -which you get for instance interacting with the dot extension- considering that we select network and then connect wallet? If some of those entities that reflect actual connection or capacity to sign transactions are provided, is the `from network` selector going to be prefilled?

mpetrunic: I think it makes sense to prefill from network if website provided api.
In theory, maybe it makes sense to only allow providing signer and we handle creating provider, at least for substrate, wdyt?

wainola: for substrate makes a lot of sense to just provide signer, since signer and provider are somewhat decouple in a way. What about ethereum? Here we give the option to pass web3Provider just because is being used on the SDK. I guess for MVP purposes makes sense to go along with that definition

packages/widget/src/interfaces/index.ts
```diff
@@ -0,0 +1,30 @@
...
+  | 'mainColor'
+  | 'secondaryColor'
+  | 'fontSize'
+  | 'borderRadius'
+  | 'borderRadiusSecondary';
+
+export type Theme = Record<ThemeVariables, string>;
+
+export interface ISygmaProtocolWidget {
+  networks?: Network;
```
mpetrunic, line 20:
```suggestion
  whitelistedNetworks?: Network[];
```

packages/widget/src/interfaces/index.ts
```diff
@@ -0,0 +1,30 @@
...
+  web3Provider?: ethers.providers.Web3Provider;
```
mpetrunic, line 21:
```suggestion
  evmProvider?: ethers.providers.Web3Provider;
```

packages/widget/src/interfaces/index.ts
```diff
@@ -0,0 +1,30 @@
...
+  apiPromise?: ApiPromise | string;
```
mpetrunic, line 22:
```suggestion
  substrateProvider?: ApiPromise | string;
```

packages/widget/src/interfaces/index.ts
```diff
@@ -0,0 +1,30 @@
...
+  signer: Signer;
+  disabled?: boolean;
+  whitelistedResource: EvmResource | SubstrateResource;
```
mpetrunic, line 25:
```suggestion
  whitelistedSourceResources?: Array<EvmResource | SubstrateResource>;
```

packages/widget/src/interfaces/index.ts
```diff
@@ -0,0 +1,30 @@
...
+  expandable: boolean;
```
mpetrunic, line 26:
```suggestion
  expandable?: boolean;
```

packages/widget/src/interfaces/index.ts
```diff
@@ -0,0 +1,30 @@
...
+  | 'mainColor'
+  | 'secondaryColor'
+  | 'fontSize'
+  | 'borderRadius'
+  | 'borderRadiusSecondary';
+
+export type Theme = Record<ThemeVariables, string>;
+
+export interface ISygmaProtocolWidget {
+  whitelistedNetworks?: Network[];
```
mpetrunic, line 20:
```suggestion
  whitelistedSourceNetworks?: Network[];
  whitelistedDestinationNetworks?: Network[];
```

packages/widget/src/interfaces/index.ts
```diff
@@ -0,0 +1,30 @@
...
+  evmProvider?: ethers.providers.Web3Provider;
```
mpetrunic, line 21: it should be https://github.com/ethers-io/ethers.js/blob/a26ff77c210660e641455150ccea6c0530f02f61/src.ts/providers/provider-browser.ts#L16

Also instead of depending on ethers package in this react package, you can just c/p this interface

mpetrunic (commented):
With those small changes, you can start building react component that wraps widget^^

mpetrunic:
> @mpetrunic since in the meantime the react package is using the widget from the workspace, do we care if the pipeline for the linter is failing because it can't find the dependency, since afaik is just linting and building that package?

nope :) fixed ci

packages/react/package.json
```diff
@@ -7,18 +7,23 @@
   "license": "LGPL-3.0-or-later",
   "type": "module",
   "dependencies": {
-    "@lit-labs/react": "2.1.2"
-  },
-  "peerDependencies": {
+    "@buildwithsygma/sygmaprotocol-widget": "workspace:^",
+    "@lit/react": "1.0.3",
     "react": ">=17",
     "react-dom": ">=17"
```
mpetrunic, line 13: shouldn't react deps be peer deps?

mpetrunic (commented):
could you also add demo react app that uses @buildwithsygma/sygmaprotocol-react-widget and change cd to deploy that react page on cloudflare?

packages/react/package.json
wainola: I had it this way, but I wasn't able to build it in order to test it

mpetrunic: you should add react as both peer (range) and dev dep (exact version) :)

wainola: done

.github/workflows/cf-deploy-react.yml
```diff
@@ -10,6 +10,7 @@ on:
       - main
     paths:
       - 'packages/react/**'
+      - 'examples/react-widget-app/**'
```
mpetrunic, line 13: lets remove paths filters, we want new deployment even if widget package is modified^^

.github/workflows/cf-deploy-react.yml
```diff
@@ -21,20 +22,24 @@ jobs:
...
-          cache: 'yarn'
           node-version: 18
       - run: corepack enable
-      - run: yarn set version stable
+      - uses: actions/setup-node@v4
+        with:
+          cache: yarn
+          node-version: 18
       - run: yarn install --immutable
+      - run: yarn build:widget
```
mpetrunic, line 32: You should build all packages to make it future proof

mpetrunic:
```suggestion
      - run: yarn build
```

.github/workflows/cf-deploy-react.yml
```diff
@@ -21,20 +22,24 @@ jobs:
...
       - run: yarn run lint:react-widget
       - run: yarn run build:react-widget
```
mpetrunic, line 34: if you build all packages, you can remove this. You should also remove lint as it's being run on the ci workflow

examples/react-widget-app/.eslintrc.cjs
```diff
@@ -0,0 +1,18 @@
+module.exports = {
```
mpetrunic, line 1: replace this and use https://github.com/ChainSafe/eslint-config?tab=readme-ov-file#frontend-react

mpetrunic (changes requested):

examples/react-widget-app/index.html
```diff
@@ -0,0 +1,12 @@
+<!doctype html>
+<html lang="en">
+  <head>
+    <meta charset="UTF-8" />
+    <meta name="viewport" content="width=device-width, initial-scale=1.0" />
+    <title>Sygma Widget React Example</title>
+  </head>
+  <body>
+    <div id="root"></div>
+    <script type="module" src="/src/main.tsx"></script>
```
sztok7, line 10: do we need this line? since app is bootstrapped to div (one line before this one) in `main.tsx`

examples/react-widget-app/.eslintrc.cjs
```diff
@@ -0,0 +1,4 @@
+
+module.exports = {
+  extends: "@chainsafe/eslint-config/frontend-react",
```
sztok7, line 3: you'll need to `yarn add` this package to make it work 😅

examples/react-widget-app/package.json
```diff
@@ -0,0 +1,32 @@
...
+  "devDependencies": {
+    "@chainsafe/eslint-config": "2.2.2",
+    "@types/react": "^18.2.43",
+    "@types/react-dom": "^18.2.17",
+    "@typescript-eslint/eslint-plugin": "^6.14.0",
+    "@typescript-eslint/parser": "^6.14.0",
+    "@vitejs/plugin-react": "^4.2.1",
+    "eslint": "^8.55.0",
+    "eslint-plugin-react-hooks": "^4.6.0",
+    "eslint-plugin-react-refresh": "^0.4.5",
```
mpetrunic, line 28: you can remove plugins, they get installed via @chainsafe/eslint-config

examples/react-widget-app/package.json
```diff
@@ -0,0 +1,32 @@
...
+    "@buildwithsygma/sygmaprotocol-react-widget": "workspace:^",
+    "react": "^18.2.0",
+    "react-dom": "^18.2.0"
+  },
+  "devDependencies": {
+    "@chainsafe/eslint-config": "2.2.2",
+    "@types/react": "^18.2.43",
+    "@types/react-dom": "^18.2.17",
+    "@typescript-eslint/eslint-plugin": "^6.14.0",
+    "@typescript-eslint/parser": "^6.14.0",
```
mpetrunic, line 24: you can remove @typescript-eslint/* packages, they get installed via @chainsafe/eslint-config

examples/react-widget-app/package.json
wainola: done

examples/react-widget-app/package.json
wainola: done

mpetrunic (approved):

sztok7 (approved):
🚀

examples/react-widget-app/package.json
```diff
@@ -0,0 +1,28 @@
...
+  "name": "@buildwithsygma/react-widget-app",
+  "private": true,
+  "version": "0.0.0",
+  "type": "module",
+  "scripts": {
+    "dev": "vite",
+    "build": "tsc && vite build",
+    "lint": "yarn run lint:types && yarn run lint:code",
+    "lint:types": "tsc -p ./tsconfig.json --noEmit",
+    "lint:code": "eslint '{src,tests}/**/*.ts'",
```
BeroBurny, line 11: It is intentional to just check `ts` files?

Issue #64, opened 2024-01-22
Network selector component


mpetrunic:
- start branch from https://github.com/sygmaprotocol/sygma-widget/pull/62
- add styling according to design
- filter source/dest network on network selection (inside WidgetController)
- render network icons

PR #70, opened 2024-01-25
feat: network selector component

Lykhoyda:
Closes: #64 
!!!! To clarify: "filter source/dest network on network selection (inside WidgetController)"


packages/widget/src/components/network-selector/network-selector.ts
```diff
@@ -16,70 +17,91 @@ type Direction = (typeof Directions)[keyof typeof Directions];
 export class NetworkSelector extends LitElement {
   static styles = styles;
 
-  @property({
-    type: Boolean
-  })
-  disabled = false;
+  @state()
+  private _isDropdownOpen = false;
```
wainola, line 21: if we use the accessor private, does it make sense to use `_something` notation? Wdyt?

packages/widget/src/components/network-selector/styles.ts
```diff
@@ -6,6 +6,7 @@ export const styles = css`
     border: 1px solid #e4e4e7;
     display: flex;
     width: 314px;
```
wainola, line 8: should we change all px to rem?

wainola (changes requested):

packages/widget/src/components/network-selector/network-selector.ts
Lykhoyda: it will work without underscore, I think it's just an old naming convention to indicate the private method in the class. But we can agree on the way how to properly write it.

packages/widget/src/components/network-selector/styles.ts
Lykhoyda: Yes, we should stick to rem as they are widely used in Figma. Also, we should avoid the exact values for the width and height and use max-width and max-height instead.

packages/widget/src/components/network-selector/network-selector.ts
wainola: we have some of them in the others components as well, but there are not private per se, just "internal" I guess. I'm not against it, but we should have some agreement. For instance `_onResourceSelected` on amount component is only being used inside the component, so it could make sense either to add `private` and mantain the naming, or add private and remove the underscore.

packages/widget/src/components/network-selector/styles.ts
wainola: I guess in the scope of the PR and the component, all units that are in px should be change them then

packages/widget/src/components/network-selector/network-selector.ts
Lykhoyda: Let's remove then private

mpetrunic:
> !!!! To clarify: "filter source/dest network on network selection (inside WidgetController)"

@Lykhoyda In sygma bridge there may not be route for every combination of source/destination networks. So when user selects source network, we should use sygma SDK to fetch available destination networks for this source network.
Unfortunately this is not yet available in sygma SDK. So for now, could you open issue for that?

packages/widget/src/components/network-selector/network-selector.ts
```diff
@@ -16,70 +17,91 @@ type Direction = (typeof Directions)[keyof typeof Directions];
...
-    this.onNetworkSelected?.(network);
+  _toggleDropdown = (): void => {
+    this._isDropdownOpen = !this._isDropdownOpen;
+  };
+
+  _selectOption(option: Domain, event: Event): void {
+    event.stopPropagation();
+    this.selectedNetwork = option;
+    this.onNetworkSelected?.(option);
+    this._isDropdownOpen = false;
```
mpetrunic, line 46: probably better to use this before callback is called. Never know how long callback execution is gonna take 🤷🏻‍♂️

mpetrunic (dismissed):
one small comment, otherwise lgtm^^

packages/widget/src/components/network-selector/network-selector.ts
Lykhoyda: Yeah, it's a good idea 👍

packages/widget/src/components/network-selector/network-selector.ts
```diff
@@ -16,70 +17,91 @@ type Direction = (typeof Directions)[keyof typeof Directions];
...
-  })
-  selected?: Domain;
+  @property({ type: Object })
+  selectedNetwork?: Domain;
 
-  @property({
-    attribute: false
-  })
+  @property({ attribute: false })
   onNetworkSelected?: (network?: Domain) => void;
```
wainola, line 33: should this property be optional?

packages/widget/src/components/network-selector/network-selector.ts
```diff
@@ -16,70 +17,91 @@ type Direction = (typeof Directions)[keyof typeof Directions];
...
-  })
-  icons = true;
+  @property({ type: Boolean })
+  disabled = false;
 
-  @property({
-    type: String
-  })
+  @property({ type: String })
   direction?: Direction;
```
wainola, line 27: what if default value of direction is from? wdty?

packages/widget/src/components/network-selector/network-selector.ts
```diff
@@ -16,70 +17,91 @@ type Direction = (typeof Directions)[keyof typeof Directions];
...
-          ${capitalize(entry.name)}
-        </option>`;
-      });
-    }
-    return html`<option selected value="">Network</option>`;
+  _renderNetworkIcon(name: string): HTMLTemplateResult {
+    return networkIconsMap[name] || networkIconsMap.default;
+  }
+
+  _renderEntries(): Generator<unknown, void> | HTMLTemplateResult {
```
wainola, line 53: why the return type is either `Generator` or `HTMLTemplateResult`?

packages/widget/src/components/network-selector/network-selector.ts
```diff
@@ -16,70 +17,91 @@ type Direction = (typeof Directions)[keyof typeof Directions];
...
+        >
+          ${this._renderNetworkIcon(network.name)}
+          <span class="networkName">${capitalize(network.name)}</span>
+        </div>
+      `
+    );
+  }
+
+  _renderTriggerContent(): HTMLTemplateResult {
+    return this.selectedNetwork
```
wainola, line 70: here you can use `choose` or `when` directives

packages/widget/src/components/network-selector/network-selector.ts
```diff
@@ -16,70 +17,91 @@ type Direction = (typeof Directions)[keyof typeof Directions];
...
+        aria-expanded="${this._isDropdownOpen ? 'true' : 'false'}"
+      >
+        <div class="dropdownTrigger">
+          <div class="selectedNetwork">${this._renderTriggerContent()}</div>
+          <div class="chevron ${this._isDropdownOpen ? 'open' : ''}">
+            ${chevronIcon}
+          </div>
+        </div>
+        <div
+          class="dropdownContent ${this._isDropdownOpen ? 'show' : ''}"
```
wainola, line 95: I think the correct way to write this should be using `${}` and backticks, although Lit allows you to define this, but to me it looks kind of weird, since you are using template strings without backticks

wainola (changes requested):

packages/widget/src/components/network-selector/network-selector.ts
Lykhoyda: In general yeah, the component without sending back the state will make no sense, so probably we should make it mandatory

packages/widget/src/components/network-selector/network-selector.ts
Lykhoyda: I think it works as we have top-level backticks from html``  ,so for such case, it's probably not required

packages/widget/src/components/network-selector/network-selector.ts
Lykhoyda: Updated with "when"

packages/widget/src/components/network-selector/network-selector.ts
Lykhoyda: I think it's better to specify explicitly in such case or render no label. I don't have a strong opinion here

mpetrunic (approved):

packages/widget/src/components/network-selector/network-selector.ts
wainola: tottaly miss out that

wainola (approved):

Issue #67, opened 2024-01-23
Resource amount input + type selector


mpetrunic:
## Implementation details

- provide styling from figma
- resources should be filtered based on src/dest networks
   - resource selector should be disabled until both networks are selected
- on resource selected, wallet balance should be presented
   - refresh balance periodically  
- display warning if input amount is greater than balance (but don't prevent transfer as we might have wrong balance) 
- implement logic for "useMax" button

PR #77, opened 2024-02-05
feat: amount input styles update

Lykhoyda:
Closes: #67 

⚠️ this part of the story is not implemented due to hardcoded data (moved to the separate issue #78 ) 

on resource selected, wallet balance should be presented
refresh balance periodically


packages/widget/src/components/amount-selector/amount-selector.ts
```diff
@@ -1,117 +1,136 @@
...
+      this.onAmountChange?.(Number.parseFloat(amount));
+    }
   };
 
-  _onResourceSelected = (event: Event): void => {
-    const { value } = event.target as HTMLOptionElement;
-    const resource = this.resources.find((n) => String(n.resourceId) == value);
+  _onResourceSelected = (option: DropdownOption): void => {
+    const resource = this.resources.find(
+      (n) => String(n.resourceId) == option.id
```
wainola, line 59:
```suggestion
      (resource) => String(resource.resourceId) == option.id
```

packages/widget/src/components/amount-selector/amount-selector.ts
```diff
@@ -1,117 +1,136 @@
...
+      this.validationMessage = 'Amount must be greater than 0';
+      return false;
+    } else if (
+      this.accountBalance &&
+      parsedAmount > Number.parseFloat(this.accountBalance)
+    ) {
+      this.validationMessage = 'Amount exceeds account balance';
+      return false;
+    } else {
+      this.validationMessage = '';
```
wainola, line 78: `validationMessage` could be null, and then we don't need to reasigned it to a string, just to null, wdty?

packages/widget/src/components/amount-selector/amount-selector.ts
```diff
@@ -1,117 +1,136 @@
...
-    attribute: false
-  })
+  @property({ attribute: false })
   onAmountChange?: (amount: number) => void;
 
   @query('.amountSelectorInput', true)
   _input!: HTMLInputElement;
 
-  useMaxBalance = (): void => {
+  @state() validationMessage: string = '';
```
wainola, line 43:
```suggestion
  @state() validationMessage: string | null = null;
```

packages/widget/src/components/amount-selector/amount-selector.ts
```diff
@@ -1,117 +1,136 @@
...
+              <dropdown-component 
+                .selectedOption=${this.preselectedToken}
+                ?disabled=${this.disabled} 
+                .onOptionSelected=${this._onResourceSelected}
+                .options=${this._normalizeOptions()}
+                >
+            </section>
+          </div>
+          <div class="errorWrapper">
+            ${this.validationMessage ? html`<div class="validationMessage">${this.validationMessage}</div>` : null}
```
wainola, line 132: we can use `when` directive here

packages/widget/src/components/internal/dropdown/dropdown.ts
```diff
@@ -0,0 +1,126 @@
...
+  _renderTriggerContent(): HTMLTemplateResult {
+    return when(
+      this._selectedOption,
+      () =>
+        html`${this._selectedOption!.icon || networkIconsMap.default}
+          <span class="optionName">
+            ${capitalize(this._selectedOption!.name)}
+          </span>`,
+      () =>
+        this.placeholder
```
wainola, line 68: you can use when here again and avoid the last call to `html`

wainola (changes requested):

packages/widget/src/components/amount-selector/amount-selector.ts
Lykhoyda: Good catch!

packages/widget/src/components/amount-selector/amount-selector.ts
Lykhoyda: We can do it with null as well 👍

packages/widget/src/components/amount-selector/amount-selector.ts
```diff
@@ -1,117 +1,141 @@
...
-    type: Number
-  })
+  @property({ type: Number })
   preselectedAmount?: number;
 
-  @property({
-    attribute: false
-  })
+  @property({ attribute: false })
   onResourceSelected?: (resource: Resource) => void;
```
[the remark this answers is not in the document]
Lykhoyda: @mpetrunic do you mean to merge it in the controller, so that we have a method like this?
```js
// Old
  onResourceSelected = (resource: Resource): void => {
    console.log('resource', resource);
    this.selectedResource = resource;
  };

  onResourceAmountChange = (amount: number): void => {
    console.log('resource amount', amount);
    this.resourceAmount = amount;
  };
  
  // New
    onResourceSelected = (resource: Resource, amount: number): void => {
    console.log('resource', resource);
    this.selectedResource = resource;
    this.resourceAmount = amount;
  };
  
```

Lykhoyda:
> * When validation fails, it probably makes sense for entire section border to be red?
> * In the design, selected token is gray pill
> * amount input font seems to be to bold/big (cc @sweetpea22 ). Maybe make by default smaller but also we could make font smaller depending on input size?
> * can we not have input border when inputting (or at least less visible)

All fixed, thank you for the review!

packages/widget/src/components/amount-selector/amount-selector.ts
```diff
@@ -1,117 +1,141 @@
...
-  @property({
-    type: Boolean
-  })
+  @property({ type: Boolean })
   disabled = false;
 
-  @property({
-    type: String
-  })
+  @property({ type: String })
```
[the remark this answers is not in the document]
Lykhoyda: @mpetrunic Sorry for the dummy question, but is it the correct understanding:
The current balance value is hardcoded to zero and fetching the balance is not yet implemented. Are you asking to add the functionality for WidgetController to fetch the balance using sygma SDK and pass this value to the "amount selector"?

packages/widget/src/components/amount-selector/amount-selector.ts
```diff
@@ -1,117 +1,147 @@
...
   _input!: HTMLInputElement;
 
-  useMaxBalance = (): void => {
-    this.preselectedAmount = Number.parseFloat(this.accountBalance!);
-    this._onInputAmountChange();
+  @state() validationMessage: string | null = null;
+
+  _useMaxBalance = (): void => {
+    this._input.value = this.accountBalance!;
+    this._validateAmount(this._input.value);
```
sztok7, line 48: validate should probably be condition of `if` clause (same as `onInputAmountChange`), because this causes some discrepancies (e.g. if user has 0 tokens and presses `Max`)

packages/widget/src/components/amount-selector/amount-selector.ts
```diff
@@ -1,117 +1,147 @@
...
-    const { value } = event.target as HTMLOptionElement;
-    const resource = this.resources.find((n) => String(n.resourceId) == value);
-    if (resource) {
-      this.onResourceSelected?.(resource);
-    }
+  _onResourceSelected = (option: DropdownOption): void => {
+    const resource = this.resources.find(
+      (resource) => String(resource.resourceId) == option.id
+    );
+    if (resource) this.onResourceSelected?.(resource);
```
sztok7, line 62: same comment as in amount selector (passing whole object as value, so it can be emitted and avoid having to `.find`). 

also, we'll need to add linter here because `===` should be used instead of `==`

sztok7 (changes requested):
also, it'd be good to close dropdown on click outside of it, as most users think of this as default UX

and to avoid this case too
image

packages/widget/src/components/amount-selector/amount-selector.ts
```diff
@@ -1,117 +1,147 @@
...
   };
 
-  renderBalance(): HTMLTemplateResult {
+  _validateAmount(amount: string): boolean {
+    const parsedAmount = Number.parseFloat(amount);
+    if (isNaN(parsedAmount) || parsedAmount <= 0) {
+      this.validationMessage = 'Amount must be greater than 0';
```
wainola, line 68: since we have error messages without uppercase in the address input -if not mistaken- we should follow the same convention here

packages/widget/src/components/amount-selector/amount-selector.ts
```diff
@@ -1,117 +1,147 @@
...
-    }
+  _onResourceSelected = (option: DropdownOption): void => {
+    const resource = this.resources.find(
+      (resource) => String(resource.resourceId) == option.id
+    );
+    if (resource) this.onResourceSelected?.(resource);
   };
 
-  renderBalance(): HTMLTemplateResult {
+  _validateAmount(amount: string): boolean {
```
wainola, line 65: does it make sense to you to move this function to the utils folder? Wdyt?

packages/widget/src/components/amount-selector/styles.ts
```diff
@@ -3,55 +3,88 @@ import { css } from 'lit';
...
+    width: 100%;
+    display: flex;
+    justify-content: space-between;
+  }
+
+  dropdown-component::part(dropdownWrapper) {
+    max-width: 8.1875rem;
+    width: 100%;
+    border-radius: 2.5rem;
+    background: var(--zinc-100, #f4f4f5);
```
wainola, line 48: why do we need the fallback color?

packages/widget/src/components/internal/dropdown/dropdown.ts
```diff
@@ -0,0 +1,128 @@
...
+  placeholder = '';
+
+  @property({ type: String })
+  label = '';
+
+  @property({ type: Array })
+  options: DropdownOption[] = [];
+
+  @property({ type: Object })
+  _selectedOption: DropdownOption | null = null;
```
wainola, line 36: is there a reason why this property is not `selectedOption`?

packages/widget/src/components/internal/dropdown/dropdown.ts
```diff
@@ -0,0 +1,128 @@
...
+  name: string;
+  icon?: HTMLTemplateResult | string;
+}
+
+@customElement('dropdown-component')
+export class Dropdown extends BaseComponent {
+  static styles = styles;
+
+  @state()
+  _isDropdownOpen = false;
```
wainola, line 21: same question here as below, since this is a internal state variable, we could just renamed as `isDropdownOpen`

packages/widget/src/components/internal/dropdown/styles.ts
```diff
@@ -0,0 +1,107 @@
...
+
+    &.open {
+      transform: rotate(180deg);
+    }
+  }
+
+  .dropdownContent {
+    display: none;
+    position: absolute;
+    background-color: white;
```
wainola, line 49: I think we have a color var for this if not mistaken unless we need this to be actually white

wainola (changes requested):

packages/widget/src/components/amount-selector/amount-selector.ts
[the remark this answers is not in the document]
mpetrunic: callback prop should have `(resource: Resource, amount: number): void` interface, so that parent component receives both resource and amount selected^^

packages/widget/src/components/amount-selector/amount-selector.ts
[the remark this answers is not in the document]
mpetrunic: I've already have branch based on this one, where AmountSelector (or rather TokenBalance controller is has attached) handles balance fetching so there is no need for parent component to pass balance.

packages/widget/src/components/internal/dropdown/dropdown.ts
```diff
@@ -0,0 +1,128 @@
...
+      () =>
+        html`${this._selectedOption!.icon || networkIconsMap.default}
+          <span class="optionName">
+            ${capitalize(this._selectedOption!.name)}
+          </span>`,
+      () =>
+        when(
+          this.placeholder,
+          () => html`<span class="placeholder">${this.placeholder}</span>`,
+          () => null
```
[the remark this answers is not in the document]
Lykhoyda: it will return undefined instead of null, but yes we can remove it

packages/widget/src/components/internal/dropdown/dropdown.ts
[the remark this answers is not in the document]
Lykhoyda: I also thought about shared, I will rename it 👍

packages/widget/src/components/internal/dropdown/dropdown.ts
```diff
@@ -0,0 +1,128 @@
...
+
+  _toggleDropdown = (): void => {
+    if (!this.disabled) {
+      this._isDropdownOpen = !this._isDropdownOpen;
+    }
+  };
+
+  _selectOption(option: DropdownOption, event: Event): void {
+    if (!this.disabled) {
+      event.stopPropagation();
```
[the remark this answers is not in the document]
Lykhoyda: Without it, it will not fire the toggle event

packages/widget/src/components/amount-selector/amount-selector.ts
```diff
@@ -1,117 +1,141 @@
...
-    attribute: false
-  })
+  @property({ attribute: false })
   onResourceSelected?: (resource: Resource) => void;
 
-  @property({
-    attribute: false
-  })
+  @property({ attribute: false })
   onAmountChange?: (amount: number) => void;
```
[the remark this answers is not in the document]
Lykhoyda: without onResourceSelected the component will still work but without retrieving the data it definitely will not make sense

packages/widget/src/components/internal/dropdown/styles.ts
```diff
@@ -0,0 +1,107 @@
...
+    width: 100%;
+    display: flex;
+    justify-content: space-between;
+    align-items: center;
+    cursor: pointer;
+    padding: 0.25rem 0;
+    box-sizing: border-box;
+
+    .optionName {
+      max-width: 4rem;
```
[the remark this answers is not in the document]
Lykhoyda: My bad, it was intended to be only for an amount selector

packages/widget/src/components/network-selector/network-selector.ts
```diff
@@ -20,90 +20,46 @@ type Direction = (typeof Directions)[keyof typeof Directions];
...
-          <span class="networkName">
-            ${capitalize(this.selectedNetwork!.name)}
-          </span>`,
-      () => html`Select Network`
+  _normalizeOptions(): DropdownOption[] {
+    return when(this.networks, () =>
+      this.networks.map((network) => ({
+        name: network.name,
+        icon: this._renderNetworkIcon(network.name)
+      }))
```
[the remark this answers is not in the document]
Lykhoyda: Great idea! thank you!

packages/widget/src/components/amount-selector/amount-selector.ts
```diff
@@ -1,117 +1,147 @@
...
 
   @query('.amountSelectorInput', true)
   _input!: HTMLInputElement;
```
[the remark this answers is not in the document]
Lykhoyda: I'm not an expert in Lit, as this code was before I started to work on the story I guessed it is "the way to do" in Lit. But I will agree the state approach is more declarative.

packages/widget/src/components/amount-selector/amount-selector.ts
Lykhoyda: Code replaced with value

packages/widget/src/components/amount-selector/styles.ts
Lykhoyda: It was copied from figma by mistake

packages/widget/src/components/amount-selector/amount-selector.ts
Lykhoyda: At least in the design it starts from uppercase, so it will be easier to follow what we have in the design
Screenshot 2024-02-07 at 15 13 05

packages/widget/src/components/amount-selector/amount-selector.ts
Lykhoyda: I don't think it's reusable now as the main purpose is to set the error message. But we can think about it later if we would need to use it in more places

packages/widget/src/components/internal/dropdown/dropdown.ts
Lykhoyda: no, i think it's a mistake

packages/widget/src/components/amount-selector/amount-selector.ts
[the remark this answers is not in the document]
sztok7: yeah, i saw that this was here before, but i tried out state approach and it works!

Lykhoyda:
> also, it'd be good to close dropdown on click outside of it, as most users think of this as default UX
> 
> and to avoid this case too image

I'm glad Marin had a quick solution for that 😅

packages/widget/src/components/amount-selector/amount-selector.ts
```diff
@@ -1,117 +1,154 @@
...
+    if (!this._validateAmount(value)) return;
 
-  _onResourceSelected = (event: Event): void => {
-    const { value } = event.target as HTMLOptionElement;
-    const resource = this.resources.find((n) => String(n.resourceId) == value);
-    if (resource) {
-      this.onResourceSelected?.(resource);
+    this.amount = value;
+    if (this.selectedResource) {
+      this.onResourceSelected?.(
```
sztok7, line 57:
```suggestion
      this.onResourceSelected(
```

packages/widget/src/components/amount-selector/amount-selector.ts
```diff
@@ -1,117 +1,154 @@
...
+      );
     }
   };
 
-  renderBalance(): HTMLTemplateResult {
+  _onResourceSelectedHandler = ({ value }: DropdownOption): void => {
+    this.selectedResource = value as Resource;
+    const amount = Number.parseFloat(this.amount!);
+
+    if (value) this.onResourceSelected?.(value as Resource, amount);
```
sztok7, line 68:
```suggestion
    if (value) this.onResourceSelected(value as Resource, amount);
```

packages/widget/src/components/common/dropdown/dropdown.ts
```diff
@@ -0,0 +1,145 @@
...
+  _selectOption(option: DropdownOption, event: Event): void {
+    if (this.disabled) return;
+
+    event.stopPropagation();
+    this.selectedOption = option;
+    this._toggleDropdown();
+    this.dispatchEvent(
+      new CustomEvent('option-selected', { detail: { option } })
+    );
+    this.onOptionSelected?.(option);
```
sztok7, line 74:
```suggestion
    this.onOptionSelected(option);
```

packages/widget/src/components/common/dropdown/dropdown.ts
```diff
@@ -0,0 +1,145 @@
...
+  };
+
+  _selectOption(option: DropdownOption, event: Event): void {
+    if (this.disabled) return;
+
+    event.stopPropagation();
+    this.selectedOption = option;
+    this._toggleDropdown();
+    this.dispatchEvent(
+      new CustomEvent('option-selected', { detail: { option } })
```
sztok7, line 72: you can just emit `value` instead of whole option, so there won't be a need for unwrapping in handler function

packages/widget/src/components/network-selector/network-selector.ts
```diff
@@ -20,90 +20,45 @@ type Direction = (typeof Directions)[keyof typeof Directions];
...
-  onNetworkSelected: (network?: Domain) => void = () => {};
+  onNetworkSelected: (option?: Domain) => void = () => {};
 
   @property({ type: Array })
   networks: Domain[] = [];
 
-  _toggleDropdown = (): void => {
-    this._isDropdownOpen = !this._isDropdownOpen;
+  _onOptionSelected = (option: DropdownOption): void => {
+    this.onNetworkSelected?.(option.value as Domain);
```
sztok7, line 36:
```suggestion
    this.onNetworkSelected(option.value as Domain);
```

mpetrunic (commented):
LGTM but you need to add tests for amount selector^^

wainola (dismissed):

wainola:
@Lykhoyda you will need to merge main to this branch since is outdated 😄

Lykhoyda:
> LGTM but you need to add tests for amount selector^^

Added the tests 👍

mpetrunic (approved):

mpetrunic (approved):

sztok7 (approved):

Issue #66, opened 2024-01-23
Widget Loader


mpetrunic:
## Implementation details

- design is pending so just create some overlay and gif,
- it should display whenever some async task is being done

PR #79, opened 2024-02-05
feat: overlay

wainola:
## Description

 Implenentation of loader overlay

## Related Issue Or Context

Closes: #66 

## Types of changes

- [ ] Documentation

## Checklist:


wainola:
@mpetrunic contrary to what I thought, since we are handling the state of the opening and close of the overlay in the widget, it doens't make sense to have event propagation from the overlay itself to signal change of the loading state. Because of that implementation of the component is pretty straigthforward. Possible improvement is to move the spiner itself to a separate component, so we can use that for the button. Wdyt?

packages/widget/src/components/overlay-component/overlay-component.ts
```diff
@@ -0,0 +1,32 @@
...
+  @property({ type: Boolean })
+  isLoading = false;
+
+  render(): HTMLTemplateResult {
+    return when(
+      this.isLoading,
+      () => html`
+      <div class="loader">
+        <div class="overlay">
+          <div class="loading-spinner"></div>
```
Lykhoyda, line 20: to be consistent with the naming it probably should be loadingSpinner

sztok7: but it'd be also good to switch to kebab case/BEM as soon as possible, agreed? definitely a separate PR material

packages/widget/src/widget.ts
```diff
@@ -73,6 +85,9 @@ class SygmaProtocolWidget extends BaseComponent {
           </section>
         </form>
         <section class="poweredBy">${sygmaLogo} Powered by Sygma</section>
+        <sygma-overlay-component
```
Lykhoyda, line 88: Should we distinguish between the internal components and the top-level ones? For example: overlay, dropdown, and button are used internally. Maybe we can name with sygma prefix the components like network or amount selector as it's used as a top-level import and the "helper" components without sygma? wdyt? 😃

Lykhoyda (dismissed):

packages/widget/src/components/overlay-component/overlay-component.ts
wainola: I have created this issue @sztok7

packages/widget/src/widget.ts
wainola: Aside from overlay, do we have another component that doesn't have the basic layout being used -either button or dropdown- coded in the same component itself? I agree with the idea but I wonder if we would need to tackle small changes in the way the use the top level import components. Just to have a better understanding

packages/widget/src/components/overlay-component/overlay-component.ts
```diff
@@ -0,0 +1,32 @@
...
+  render(): HTMLTemplateResult {
+    return when(
+      this.isLoading,
+      () => html`
+      <div class="loader">
+        <div class="overlay">
+          <div class="loadingSpinner"></div>
+          </div>
+        </div>
+      </div>`
```
sztok7, line 23: 3 opening tags and 4 closing?

packages/widget/src/components/overlay-component/overlay-component.ts
```diff
@@ -0,0 +1,32 @@
...
+@customElement('sygma-overlay-component')
+export class OverlayComponent extends LitElement {
+  static styles = styles;
+
+  @property({ type: Boolean })
+  isLoading = false;
+
+  render(): HTMLTemplateResult {
+    return when(
+      this.isLoading,
```
sztok7, line 16: IMHO it's much better not to render/mount component, instead of passing boolean prop which says if component should render anything.

sztok7 (changes requested):

packages/widget/src/widget.ts
mpetrunic: @Lykhoyda even internal components exists in global namespace. In case our widget is used in another app, that registered "overlay" component with the same name, we could clash and our widget might render their component instead of ours.

packages/widget/src/components/overlay-component/styles.ts
```diff
@@ -0,0 +1,35 @@
...
+    left: 0;
+    width: 100%;
+    height: 100%;
+    z-index: 9999;
+    display: flex;
+    flex-direction: column;
+    justify-content: center;
+    align-items: center;
+    background-color: rgba(255, 255, 255, 0.7);
+  }
```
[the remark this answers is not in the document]
mpetrunic: Q: isn't this gonna render overlay over entire page and not just widget?

Wouldn't it make more sense for overlay component to only overlay it's children components only?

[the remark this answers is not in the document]
sztok7: i think it looks for closest parent with `position: relative;`, so we gucci

mpetrunic:
ugh, I think you need to fix those conflicts 🙈

PR #90, opened 2024-02-12
feat: add buttons states

### LINKED ISSUE #76: Transfer Action Button Component
## Implementation details

We need button that has multiple states, including loading and disabled state.

States:
image
image
image
image
image

Lykhoyda:
Closes #76 
- Add button component
- moved components to the common folder
- update ESlint


packages/widget/src/components/common/buttons/button.styles.ts
```diff
@@ -0,0 +1,63 @@
...
+    }
+
+    &:active,
+    &:hover {
+      filter: brightness(140%);
+    }
+
+    &.disabled {
+      color: var(--zinc-400, #a1a1aa);
+      background: var(--zinc-200, #e4e4e7);
```
wainola, line 42: do we need the fallback color?

packages/widget/src/components/common/buttons/button.styles.ts
```diff
@@ -0,0 +1,63 @@
...
+
+export const buttonStyles = css`
+  .button {
+    width: 100%;
+    display: flex;
+    align-items: center;
+    justify-content: center;
+    padding: 0.75rem 1.25rem;
+    border-radius: 1rem;
+    background: var(--zinc-900, #18181b);
```
wainola, line 11: do we need the fallback colors?

packages/widget/src/components/common/buttons/button.styles.ts
```diff
@@ -0,0 +1,63 @@
...
+      color: var(--zinc-400, #a1a1aa);
+      background: var(--zinc-200, #e4e4e7);
+
+      &:hover {
+        filter: brightness(80%);
+      }
+    }
+
+    &.loading {
+      color: var(--zinc-200, #e4e4e7);
```
wainola, line 50: same question as above

packages/widget/src/components/common/buttons/button.ts
```diff
@@ -0,0 +1,55 @@
...
+export class Button extends BaseComponent {
+  static styles = buttonStyles;
+
+  @property({})
+  text: string = '';
+
+  @property({ type: Object })
+  onClick: () => void = () => {};
+
+  @property({ type: Boolean, reflect: true })
```
wainola, line 21: I wonder why the reflect true? 🤔

wainola (changes requested):

packages/widget/src/components/address-input/address-input.ts
```diff
@@ -5,7 +5,8 @@ import { ifDefined } from 'lit/directives/if-defined.js';
 import { Network } from '@buildwithsygma/sygma-sdk-core';
 import { when } from 'lit/directives/when.js';
 import { validateAddress } from '../../utils';
-import { BaseComponent } from '../base-component/base-component';
+
+import { BaseComponent } from '../common/base-component/base-component';
```
wainola, line 9: can we use named exports from some `index.ts` file? In doing so we avoid having `../common/base-component/base-component` and hence we would have `../common/base-component` as the url for the import? Wdty?

packages/widget/src/components/common/buttons/button.ts
```diff
@@ -0,0 +1,55 @@
...
+    const buttonClasses = classMap({
+      button: true,
+      disabled: this.disabled,
+      loading: this.isLoading
+    });
+
+    return html`<button
+      type="button"
+      ?disabled="${this.disabled}"
+      @click=${ifDefined(
```
wainola, line 41: did the linter ask you to add the `ifDefined` directive? 🤔

packages/widget/src/components/common/buttons/button.ts
```diff
@@ -0,0 +1,55 @@
...
+import { customElement, property } from 'lit/decorators.js';
+import { html } from 'lit';
+import { classMap } from 'lit/directives/class-map.js';
+import { when } from 'lit/directives/when.js';
+import { ifDefined } from 'lit/directives/if-defined.js';
+import { BaseComponent } from '../base-component/base-component';
+import { loaderIcon } from '../../../assets';
+import { buttonStyles } from './button.styles';
+
+@customElement('sygma-button')
```
wainola, line 11: perhaps we should use a more semantic name, since this is not a plain button, but rather one that reflect to loading state

packages/widget/tests/unit/components/common/buttons/button.test.ts
```diff
@@ -0,0 +1,65 @@
...
+    const el = document.createElement('sygma-button');
+    assert.instanceOf(el, Button);
+  });
+
+  it('displays the button text', async () => {
+    const el = await fixture<Button>(
+      html`<sygma-button text="Click me"></sygma-button>`
+    );
+    assert.equal(el.text, 'Click me');
+    assert.include(el.shadowRoot?.textContent || '', 'Click me');
```
wainola, line 21: why the or clause on the assertion?

wainola (changes requested):

packages/widget/src/components/common/buttons/button.ts
mpetrunic: sygma-action-button?

packages/widget/src/components/common/buttons/button.styles.ts
Lykhoyda: nope 😅

packages/widget/src/components/common/buttons/button.ts
Lykhoyda: I thought about styling the attribute similar to https://lit.dev/docs/components/properties/#:~:text=Permalink%20to%20%E2%80%9CEnabling,%23

But used the class in the end. So we can remove it to not overcomplicate the code

packages/widget/src/components/address-input/address-input.ts
Lykhoyda: It's a good idea and I tried to do so. But somehow re-export for abstract classes doesn't work. Looks like this issue has existed for quite a long time

packages/widget/src/components/common/buttons/button.ts
Lykhoyda: removed, actually lint ask to remove it

packages/widget/src/components/amount-selector/amount-selector.ts
```diff
@@ -132,20 +131,20 @@ export class AmountSelector extends BaseComponent {
               class="amountSelectorInput"
               placeholder="0.000"
               @change=${this._onInputAmountChangeHandler}
-              value=${this.amount || ifDefined(this.preselectedAmount)}
+              value=${this.amount ||
+              ifDefined(this.preselectedAmount).toString()}
```
mpetrunic, line 135: Can you rollback lint changes? It will cause a massive amount of conflicts, it should be in separate PR.

mpetrunic (changes requested):

Issue #86, opened 2024-02-12
Externalize walletconnect configuration


mpetrunic:
## Implementation details
Currently wallet connect configuration is hardcoded, make sure widget user can pass wallet connect options. If at least projected for wallet connect is submitted, enable wallet connect!

PR #94, opened 2024-02-14
feat: externalize wallet connect connection through props

wainola:
## Related Issue Or Context

Closes: #86 

## Types of changes

- [ ] Documentation

## Checklist:


wainola:
@mpetrunic can you give it a first look and provide feedback pls?

packages/widget/src/interfaces/index.ts
```diff
@@ -22,6 +22,11 @@ export interface Eip1193Provider {
   }): Promise<unknown>;
 }
 
+export type WalletConnectOptions = {
```
mpetrunic, line 25: don't define your options, use types from web3-onboard [package](https://github.com/blocknative/web3-onboard/blob/develop/packages/walletconnect/src/types.ts#L3)

You could also expose: https://github.com/blocknative/web3-onboard/blob/develop/packages/core/src/types.ts#L33

mpetrunic (commented):
It probably makes sense to define context for global widget options instead of passing them to each component

packages/widget/src/interfaces/index.ts
wainola: this is weird, I tried to import them but I couldn't do it. I'm going to recheck again

wainola:
> It probably makes sense to define context for global widget options instead of passing them to each component

This is a great idea

packages/widget/src/interfaces/index.ts
mpetrunic: You can always extract it using typeof+ArgumentsType

wainola:
> It probably makes sense to define context for global widget options instead of passing them to each component

@mpetrunic seeking clarification here: do you mean context for all the widget options that we currently defined as property or just the ones concerning the wallet connection? Because if this is true for the former, wouldn't make sense to tackle that on a different PR with it's own issue?

mpetrunic:
> > It probably makes sense to define context for global widget options instead of passing them to each component
> 
> @mpetrunic seeking clarification here: do you mean context for all the widget options that we currently defined as properties or just the ones concerning the wallet connection? Because if this is true for the former, wouldn't make sense to tackle that on a different PR with it's own issue?

I guess, for now just theme options and wallet options^^

packages/widget/src/components/common/buttons/connect-wallet.ts
```diff
@@ -47,10 +46,13 @@ export class ConnectWalletButton extends BaseComponent {
   }
 
   private onConnectClicked = (): void => {
+    const walletOptions = {
```
mpetrunic, line 49: why creating new object? it should work if you just pass configContext, no?

packages/widget/src/context/config.ts
```diff
@@ -0,0 +1,72 @@
...
+  Symbol('sygma-config-context')
+);
+
+declare global {
+  interface HTMLElementEventMap {
+    configUpdate: ConfigUpdateEvent;
+  }
+}
+
+export class ConfigUpdateEvent extends CustomEvent<Partial<ConfigContext>> {
```
mpetrunic, line 26: Does it make sense for child component to update config?

packages/widget/src/context/config.ts
```diff
@@ -0,0 +1,72 @@
...
+  private configContext: ConfigContext = {};
+
+  @property({ attribute: false, type: Object })
+  walletConnectOptions?: WalletConnectOptions;
+
+  @property({ attribute: false, type: Object })
+  appMetadata?: AppMetadata;
+
+  @property({ attribute: false, type: Object })
+  theme?: Theme;
```
mpetrunic, line 44: you should hook into property update and update context. That way if widget user updates theme it gets propagated (for example switching between light and dark theme).

Example of listening on property updates: https://github.com/sygmaprotocol/sygma-widget/blob/main/packages/widget/src/components/common/buttons/connect-wallet.ts#L42

Alternative is to do it in render but it seems like bad practice

packages/widget/src/controllers/wallet-manager/manager.ts
```diff
@@ -90,18 +98,27 @@ export class WalletController implements ReactiveController {
...
-        injected,
+    const walletSetup = [injected];
+
+    if (options?.walletConnectOptions?.projectId) {
+      walletSetup.push(
         walletConnectModule({
-          projectId: '2f5a3439ef861e2a3959d85afcd32d06',
-          dappUrl: options?.dappUrl
+          projectId: options.walletConnectOptions?.projectId,
+          dappUrl: options.walletConnectOptions?.dappUrl
```
mpetrunic, line 114: you should replace this with:
```
walletConnectModule(options.walletConnectOptions)
```

packages/widget/src/context/config.ts
wainola: I think not, I was thinking about theming here mostly, but those properties are static.

packages/widget/src/context/config.ts
wainola: @mpetrunic implemented. I thought I would need to use force for this, since there is an update on an object, but here we are not dealing with deep property update

wainola:
btw @mpetrunic I think we might need to expose the `ISygmaProtocolWidget` from the react package, because with the react example, there is no type inferring when for instance defining a dynamic theme with setState

mpetrunic (approved):

mpetrunic:
> btw @mpetrunic I think we might need to expose the `ISygmaProtocolWidget` from the react package, because with the react example, there is no type inferring when for instance defining a dynamic theme with setState

Thats fine

sztok7 (approved):
LGTM

Issue #97, opened 2024-02-16
Expose environment property to widget user


mpetrunic:
## Implementation details

Currently, the environment is hardcoded to the testnet. We should allow users to set it themselves when using the widget. It should default to mainnet. Our Cloudflare deployment should override and use TESTNET.

PR #95, opened 2024-02-15
feat: evm token transfer logic

mpetrunic:
## Description

A lot but tldr connecting logic for making fungible transfer from evm side.

## Related Issue Or Context

Closes: #92 

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation

## Checklist:

- [ ] I have commented my code, particularly in hard-to-understand areas.
- [ ] I have ensured that all acceptance criteria (or expected behavior) from issue are met
- [ ] I have updated the documentation locally and in sygma-docs.
- [ ] I have added tests to cover my changes.
- [ ] I have ensured that all the checks are passing and green, I've signed the CLA bot

PR #104, opened 2024-02-19
feat: allow bridge environment configuration

mpetrunic:
## Description
depends on #95 

## Related Issue Or Context

Closes: #97 

## Types of changes

- [ ] Documentation

## Checklist:


packages/widget/src/vite-env.d.ts
```diff
@@ -0,0 +1,10 @@
+/// <reference types="vite/client" />
+
+interface ImportMetaEnv {
+  readonly VITE_BRIDGE_ENV: string;
```
sztok7, line 4: why not `Environment` instead of `string`? this way those type assertions in the next file become redundant + we have more type safety

sztok7 (changes requested):

packages/widget/src/controllers/transfers/fungible-token-transfer.ts
```diff
@@ -83,7 +83,7 @@ export class FungibleTokenTransferController implements ReactiveController {
   async init(env: Environment): Promise<void> {
     this.host.requestUpdate();
     this.env = env;
-    await this.config.init(1, env);
+    await this.config.init(1, this.env);
```
wainola, line 86: why do we have chain id for mainnet hardcoded?

wainola (changes requested):

packages/widget/src/vite-env.d.ts
mpetrunic: Can you guarantee env will always be enum value?
 I'm usually quite against external API relying on typescript types as you might have a javascript user that can submit anything and you should validate external inputs (even from developer)

packages/widget/src/controllers/transfers/fungible-token-transfer.ts
mpetrunic: because it's not used in our code and should be removed in SDK because it doesn't make sense 🤷🏻‍♂️

packages/widget/src/vite-env.d.ts
sztok7: i'd definitely leave the validation in `connectedCallback`, but yeah, makes sense to leave input as `string` since it can be anything. but then, we can immediately (when setting `@property environment`) check if it is an enum value and fallback to mainnet + throw (in `connectedCallback`), so prop is always of type `Environment`? 🤷

sztok7 (approved):

wainola (approved):

PR #125, opened 2024-02-29
feat: preselected destination wallet

### LINKED ISSUE #102: Preselect destination wallet to be connected wallet
## Implementation details

When transferring between same type of network, we should autoinput connected wallet address as destination address.

Lykhoyda:
The logic will work only for EVM networks as substrate-based networks will have different prefixes.

Closes #102


wainola:
@mpetrunic will this PR make [this other](https://github.com/sygmaprotocol/sygma-widget/pull/128) PR irrelevant? Since whenever we change the network we are cleaning for diff type and appending for same type 🤔

packages/widget/src/components/transfer/fungible/fungible-token-transfer.ts
```diff
@@ -76,6 +76,19 @@ export class FungibleTokenTransfer extends BaseComponent {
     }
   };
 
+  private isSourceAndDestinationNetworkHasEqualType(): boolean {
```
wainola, line 79:
```suggestion
  private sourceAndDestinationNetworkHasEqualType(): boolean {
```

wainola (changes requested):

wainola:
@Lykhoyda there is a small issue. If you connect wallet, select diff type of network on dest network and paste a valid substrate address, validation is going to complain saying that is a invalid ethereum address 🙈

Lykhoyda:
> The logic will work only for EVM networks as substrate-based networks will have different prefixes.
> 
> Closes #102

@wainola yeah, it's true. I created a new PR that I will merge to the main and synced with your previous solution https://github.com/sygmaprotocol/sygma-widget/pull/128

Issue #111, opened 2024-02-26
Switching between networks ( from and to) it fails to retrieve network events


LyonSsS:
Severity: High | Priority: High

**Note**: When switching between test networks, a console error is generated and widget fails to retrieve data from chains.

![image](https://github.com/sygmaprotocol/sygma-widget/assets/30255634/2eeca7c8-cdc0-4e29-a3e5-96803b4a8840)

Exp: To have the option to switch network. When switching network when switching source. The console errors will also be solved.

LyonSsS:
Fixed with https://github.com/sygmaprotocol/sygma-widget/pull/127

LyonSsS:
Fixed with https://github.com/sygmaprotocol/sygma-widget/pull/127 - caused by wallet not prompting to the new network

Issue #119, opened 2024-02-26
Allow incompatible routes to be selected from UI


LyonSsS:
Severity: Medium | Priority: Medium
**Context**:  After switching to a From network that allows multiple To destinations, then the User can select an incompatible From - To mix. 

**STR**:
1. Select From network as "Sepolia"
2. Select To destination as "Mumbai"
3. Select From network as "Cronos" 

**Exp**: The To destination to be updated only to "Sepolia" as compatible match.
**Act**: The To destination were not updated to match the correct route. 

![image](https://github.com/sygmaprotocol/sygma-widget/assets/30255634/9200777b-1c5f-4dce-81fd-b804fa7c97a8)

MakMuftic:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Sygma-Protocol-62f126a54b7d5a55203732f0/issues/sygmaprotocol/sygma-widget/119?planning-poker) with Zenhub @Lykhoyda @mpetrunic @wainola

wainola:
I think this issue could be linked to #112. CC: @MakMuftic @mpetrunic

LyonSsS:
Fixed with https://github.com/sygmaprotocol/sygma-widget/pull/127

LyonSsS:
Retested - fixed with https://github.com/sygmaprotocol/sygma-widget/pull/127

Issue #118, opened 2024-02-26
Refresh MAX and Amount to transfer field when changing Source


LyonSsS:
Severity: Low | Priority: Low

**Context**: After selecting a relationship Source-Dest and filling in some "Amount to transfer", if the user then switches to another source, the context of Max and Amount to transfer is not refreshed. 

**STR**:
1. Select From  as "Mumbai" and connect metmask wallet to the correct network.
2. Select To as "Sepolia" and then select ERC20RLTest token.
3. After balance is updated, input an amount <= Max value for the "Amount to transfer" field.
4. Change From value now to "Cronos".

**Exp**: The context to be resseted for the token type to be transferred, max amount, and amount to be transferred. 
**Act**: The only context that is resseted is for the token type. 
![image](https://github.com/sygmaprotocol/sygma-widget/assets/30255634/718a44d5-8a6f-497f-81b2-f652a26ade6c)

wainola:
this issue is related to #111

MakMuftic:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Sygma-Protocol-62f126a54b7d5a55203732f0/issues/sygmaprotocol/sygma-widget/118?planning-poker) with Zenhub @Lykhoyda @mpetrunic @wainola

LyonSsS:
Tested based on fix https://github.com/sygmaprotocol/sygma-widget/pull/127 
The "Amount to transfer", still persist
<img width="848" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBekd3QXc9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--25a200042fe8d29d47366daff5b51fdb4fe37cdb/Amount_To_Transfer.gif" alt="Amount_To_Transfer.gif" />

LyonSsS:
Retested on https://github.com/sygmaprotocol/sygma-widget/pull/127 - the amount issue still persists

LyonSsS:
retested and fixed https://github.com/sygmaprotocol/sygma-widget/pull/127

Issue #112, opened 2024-02-26
Existing capability to select From and To the same network


LyonSsS:
Severity: Low | Priority: Medium

**Context**: Currently there is a possibility to select the same network value for the FROM and TO drop down value. 
![image](https://github.com/sygmaprotocol/sygma-widget/assets/30255634/96842b67-6956-4397-96bf-4d4a88739f67)

LyonSsS:
Tested and fixed with https://github.com/sygmaprotocol/sygma-widget/pull/127

PR #127, opened 2024-02-29
fix: switch networks and misc fixes

wainola:
## Related Issue Or Context

Closes: #111 #119 #118 #112 

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] Documentation

## Checklist:

### DIFF
--- a/packages/widget/src/components/amount-selector/amount-selector.ts
+++ b/packages/widget/src/components/amount-selector/amount-selector.ts
@@ -1,5 +1,6 @@
 import type { Resource } from '@buildwithsygma/sygma-sdk-core';
-import { utils, BigNumber } from 'ethers';
+import type { BigNumber } from 'ethers';
+import { utils } from 'ethers';
 import type { HTMLTemplateResult, PropertyDeclaration } from 'lit';
 import { html } from 'lit';
 import { customElement, property, state } from 'lit/decorators.js';
@@ -39,7 +40,7 @@ export class AmountSelector extends BaseComponent {
   onResourceSelected: (resource: Resource, amount: BigNumber) => void =
     () => {};
 
-  @property({ type: Object }) selectedResource: Resource | null = null;
+  @state() selectedResource: Resource | null = null;
   @state() validationMessage: string | null = null;
   @state() amount: number = 0;
 
@@ -78,17 +79,15 @@ export class AmountSelector extends BaseComponent {
     }
   }
 
-  _onResourceSelectedHandler = ({ value }: DropdownOption<Resource>): void => {
-    this.selectedResource = value;
-    this.amount = 0;
-    this.tokenBalanceController.startBalanceUpdates(value);
-    this.onResourceSelected(
-      this.selectedResource,
-      utils.parseUnits(
-        BigNumber.from(this.amount).toString(),
-        this.selectedResource.decimals ?? DEFAULT_ETH_DECIMALS
-      )
-    );
+  _onResourceSelectedHandler = (option?: DropdownOption<Resource>): void => {
+    if (option) {
+      this.selectedResource = option.value;
+      this.amount = 0;
+      this.tokenBalanceController.startBalanceUpdates(this.selectedResource);
+    } else {
+      this.selectedResource = null;
+      this.tokenBalanceController.resetBalance();
+    }
   };
 
   _validateAmount(amount: string): boolean {
@@ -173,11 +172,6 @@ export class AmountSelector extends BaseComponent {
             />
             <section class="selectorSection">
               <dropdown-component
-                .preselectedOption=${this._normalizeOptions().filter(
-                  (o) =>
-                    o.id === this.preselectedToken ||
-                    o.name === this.preselectedToken
-                )}
                 ?disabled=${this.disabled}
                 .onOptionSelected=${this._onResourceSelectedHandler}
                 .options=${this._normalizeOptions()}
--- a/packages/widget/src/components/amount-selector/amount-selector.ts
+++ b/packages/widget/src/components/amount-selector/amount-selector.ts
@@ -5,6 +5,7 @@ import { html } from 'lit';
 import { customElement, property, state } from 'lit/decorators.js';
 import { classMap } from 'lit/directives/class-map.js';
 import { when } from 'lit/directives/when.js';
+import type { PropertyValues } from '@lit/reactive-element';
 import { networkIconsMap } from '../../assets';
 import { DEFAULT_ETH_DECIMALS } from '../../constants';
 import {
@@ -21,8 +22,7 @@ export class AmountSelector extends BaseComponent {
   static styles = styles;
 
   @property({
-    type: Array,
-    hasChanged: (n, o) => n !== o
+    type: Array
   })
   resources: Resource[] = [];
 
@@ -135,6 +135,12 @@ export class AmountSelector extends BaseComponent {
         value: entry
       }))
     );
+  }
+
+  updated(changedProperties: PropertyValues): void {
+    if (changedProperties.has('resources')) {
+      this.tokenBalanceController.resetBalance();
+    }
   }
 
   render(): HTMLTemplateResult {
--- a/packages/widget/src/components/amount-selector/amount-selector.ts
+++ b/packages/widget/src/components/amount-selector/amount-selector.ts
@@ -1,5 +1,5 @@
 import type { Resource } from '@buildwithsygma/sygma-sdk-core';
-import { utils, type BigNumber } from 'ethers';
+import { utils, BigNumber } from 'ethers';
 import type { HTMLTemplateResult, PropertyDeclaration } from 'lit';
 import { html } from 'lit';
 import { customElement, property, state } from 'lit/decorators.js';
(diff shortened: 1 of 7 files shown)


packages/widget/src/components/transfer/fungible/fungible-token-transfer.ts
```diff
@@ -31,6 +31,9 @@ export class FungibleTokenTransfer extends BaseComponent {
   @property({ type: Object })
   onSourceNetworkSelected?: (domain: Domain) => void;
 
+  @state()
+  network: Domain | undefined;
```
mpetrunic, line 35: why do you need this? You have that in transferController

packages/widget/src/components/transfer/fungible/fungible-token-transfer.ts
```diff
@@ -85,6 +88,14 @@ export class FungibleTokenTransfer extends BaseComponent {
     )!;
   }
 
+  protected updated(changedProperties: PropertyValues): void {
+    if (changedProperties.has('network')) {
+      void this.walletController.switchChain(this.network?.chainId as number);
```
mpetrunic, line 93: - please move this logic into `onNetworkSelected` handler of source network selector (L124)
- You should switch network only if network is not undefined

packages/widget/src/components/transfer/fungible/fungible-token-transfer.ts
```diff
@@ -85,6 +88,14 @@ export class FungibleTokenTransfer extends BaseComponent {
...
+      // reset the selector upon changing the network
+      this.transferController.onDestinationNetworkSelected(undefined);
```
mpetrunic, line 95: this should be inside transferController. It would also be better UX if you reset selector only if currently selected destination is not supported route

packages/widget/src/components/amount-selector/amount-selector.ts
```diff
@@ -137,6 +137,12 @@ export class AmountSelector extends BaseComponent {
     );
   }
 
+  updated(changedProperties: PropertyValues): void {
+    if (changedProperties.has('resources')) {
```
mpetrunic, line 141: shouldn't this look for change in selectedResource?

mpetrunic (changes requested):

packages/widget/src/components/transfer/fungible/fungible-token-transfer.ts
wainola: to use it in updated lifecycle method

mpetrunic: Once you do https://github.com/sygmaprotocol/sygma-widget/pull/127#discussion_r1512402506 you won't need lifecycle method

wainola: did it

packages/widget/src/components/amount-selector/amount-selector.ts
wainola: if I change network, the resource is not the same, and whenever we change networks the resources are updated

mpetrunic: But selected resource should update to `undefined` if previously selected resource is not on the updated resource list^^

wainola: seeking clarification here:

- expected behaviour is if you change source but on your dest network remains the same and the route is valid, like:

`from: cronos => mumbai`
`to: sepolia`

In this case, should we reset the balance to 0 and then you select the resource to fetch the balance?

- what happens when you want switch to an invalid route, like from sepolia/mumbai to cronos/mumbai. Currently with the changes here, we are reseting the destination dropdown since that route doesn't exist, should we reset the balance to zero?

I'm asking this because as user I would expect, since there is no token selected, that if I change the network and the token gets de-selected, the balance goes to 0 😄

mpetrunic: - if you are changing source network to supported route
   - reset balance to zero
   - reset selected token if doesn't exist on the list of resources
- if you are changing source network to unsupported route
   - reset balance and selected token and destination network (network selector component should handle that because selected dest network is not on list of networks)

wainola: @mpetrunic I think the flow is working as expected, it took me some time to solve this "detail". Let me know if you have more feedback

mpetrunic:
@wainola Here is PR to address the problem causing bugs -> we weren't handling/notifying parent components that selected resource is undefined because out handlers had incorrect signatures

https://github.com/sygmaprotocol/sygma-widget/pull/138

packages/widget/src/controllers/transfers/fungible-token-transfer.ts
```diff
@@ -245,13 +253,49 @@ export class FungibleTokenTransferController implements ReactiveController {
...
-    //unselect destination if equal to source network or isn't in list of available destination networks
-    if (
-      this.destinationNetwork?.id === sourceNetwork.id ||
-      !this.supportedDestinationNetworks.includes(this.destinationNetwork!)
-    ) {
+
+    // unselect destination if equal to source network or isn't in list of available destination networks
+    if (this.destinationNetwork?.id === sourceNetwork.id) {
       this.destinationNetwork = undefined;
+      this.selectedResource = undefined;
```
mpetrunic, line 260: you should also empty supportedResources^^

packages/widget/src/controllers/transfers/fungible-token-transfer.ts
```diff
@@ -245,13 +253,49 @@ export class FungibleTokenTransferController implements ReactiveController {
...
+    }
+
+    // imposible route source -> source
+    const isSourceOnSuportedDestinations =
+      this.supportedDestinationNetworks.some(
+        (destinationDomain) =>
+          destinationDomain.chainId === this.sourceNetwork?.chainId
+      );
+
+    if (isSourceOnSuportedDestinations) {
```
mpetrunic, line 270: Why not just update filter on L242 instead of copying it here slightly modified?

packages/widget/src/controllers/transfers/fungible-token-transfer.ts
```diff
@@ -245,13 +253,49 @@ export class FungibleTokenTransferController implements ReactiveController {
...
+        .get(sourceNetwork.chainId)!
+        .filter(
+          (route) =>
+            route.toDomain.chainId !== sourceNetwork.chainId &&
+            !this.supportedDestinationNetworks.includes(route.toDomain)
+        )
+        .map((route) => route.toDomain);
     }
+
+    const hasSelectedResourceChange =
```
mpetrunic, line 281: Don't think you need this anymore, it will get automatically handled by amountSelector component now

mpetrunic (changes requested):

packages/widget/src/components/amount-selector/amount-selector.ts
```diff
@@ -82,6 +82,13 @@ export class AmountSelector extends BaseComponent {
     this.selectedResource = value;
     this.amount = 0;
     this.tokenBalanceController.startBalanceUpdates(value);
+    this.onResourceSelected(
+      this.selectedResource,
+      utils.parseUnits(
+        BigNumber.from(this.amount).toString(),
```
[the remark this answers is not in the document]
wainola: for some reason linter wasn't recognizing the BigNumber but I might missread that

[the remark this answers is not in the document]
wainola: I think you solve this with the last Pr merge to this

packages/widget/src/components/amount-selector/amount-selector.ts
```diff
@@ -137,6 +143,14 @@ export class AmountSelector extends BaseComponent {
     );
   }
 
+  updated(changedProperties: PropertyValues): void {
```
mpetrunic, line 146:
```suggestion
  updated(changedProperties: PropertyValues<this>): void {
```

packages/widget/src/controllers/transfers/fungible-token-transfer.ts
```diff
@@ -225,17 +225,49 @@ export class FungibleTokenTransferController implements ReactiveController {
...
-      this.supportedDestinationNetworks = this.routesCache
-        .get(sourceNetwork.chainId)!
-        .filter(
-          (route) =>
-            route.toDomain.chainId !== sourceNetwork.chainId &&
-            !this.supportedDestinationNetworks.includes(route.toDomain)
-        )
+      this.supportedDestinationNetworks = routes
+        .filter((route) => {
+          console.log(
```
mpetrunic, line 244: remove pls

wainola: removed

Lykhoyda (approved):

mpetrunic (approved):

mpetrunic (approved):

LyonSsS (approved):
Tested each related issue from the ticket and they work as intended.

Lykhoyda (approved):

PR #129, opened 2024-03-05
feat: set destination address for EVM

Lykhoyda:
The logic will work only for EVM networks as substrate-based networks will have different prefixes.

Closes https://github.com/sygmaprotocol/sygma-widget/issues/102

### DIFF
--- a/packages/widget/src/components/transfer/fungible/fungible-token-transfer.ts
+++ b/packages/widget/src/components/transfer/fungible/fungible-token-transfer.ts
@@ -76,25 +76,6 @@ export class FungibleTokenTransfer extends BaseComponent {
     }
   };
 
-  private sourceAndDestinationNetworkHasEqualType(): boolean {
-    return (
-      this.transferController.sourceNetwork?.type ===
-      this.transferController.destinationNetwork?.type
-    );
-  }
-
-  private getSenderDefaultDestinationAddress(): string {
-    if (
-      !this.transferController.sourceNetwork ||
-      !this.transferController.destinationNetwork
-    )
-      return '';
-
-    return this.sourceAndDestinationNetworkHasEqualType()
-      ? this.walletController.walletContext.value?.evmWallet?.address || ''
-      : this.transferController.destinatonAddress || '';
-  }
-
   render(): HTMLTemplateResult {
     const state = this.transferController.getTransferState();
     return choose(
@@ -161,7 +142,7 @@ export class FungibleTokenTransfer extends BaseComponent {
       </section>
       <section>
         <sygma-address-input
-          .address=${this.getSenderDefaultDestinationAddress()}
+          .address=${this.transferController.defaultDestinationAddress}
           .onAddressChange=${this.transferController.onDestinationAddressChange}
           .networkType=${this.transferController.destinationNetwork?.type}
         >
--- a/packages/widget/src/components/transfer/fungible/fungible-token-transfer.ts
+++ b/packages/widget/src/components/transfer/fungible/fungible-token-transfer.ts
@@ -84,6 +84,12 @@ export class FungibleTokenTransfer extends BaseComponent {
   }
 
   private getSenderDefaultDestinationAddress(): string {
+    if (
+      !this.transferController.sourceNetwork ||
+      !this.transferController.destinationNetwork
+    )
+      return '';
+
     return this.sourceAndDestinationNetworkHasEqualType()
       ? this.walletController.walletContext.value?.evmWallet?.address || ''
       : this.transferController.destinatonAddress || '';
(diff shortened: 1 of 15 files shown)


wainola (approved):

packages/widget/src/components/transfer/fungible/fungible-token-transfer.ts
```diff
@@ -76,6 +76,19 @@ export class FungibleTokenTransfer extends BaseComponent {
     }
   };
 
+  private sourceAndDestinationNetworkHasEqualType(): boolean {
+    return (
+      this.transferController.sourceNetwork?.type ===
+      this.transferController.destinationNetwork?.type
```
mpetrunic, line 82: You should check if at least one network is defined otherwise when starting new transfer they are both undefined and wallet address get's set.

Truth be told, there is a bug there because when starting new transfer we shouldn't reset source network

mpetrunic (changes requested):
When starting new transfer

packages/widget/src/components/transfer/fungible/fungible-token-transfer.ts
```diff
@@ -142,7 +161,7 @@ export class FungibleTokenTransfer extends BaseComponent {
       </section>
       <section>
         <sygma-address-input
-          .address=${this.transferController.destinatonAddress}
+          .address=${this.getSenderDefaultDestinationAddress()}
```
mpetrunic, line 164: This isn't triggering an update in transferController. So I've ended up in this state:
image

mpetrunic (changes requested):

packages/widget/src/components/transfer/fungible/fungible-token-transfer.ts
sztok7: @Lykhoyda maybe it'd make more sense to put all that logic in `transferController` instead of having it inside a component? also, that'd would probably autofix this bug with not updating state :)

Lykhoyda: @sztok7 good idea! Take a look please if I miss something

packages/widget/src/controllers/transfers/fungible-token-transfer.ts
```diff
@@ -40,6 +40,7 @@ export class FungibleTokenTransferController implements ReactiveController {
   public selectedResource?: Resource;
   public resourceAmount: BigNumber = BigNumber.from(0);
   public destinatonAddress: string = '';
+  public defaultDestinationAddress: string = '';
```
mpetrunic, line 43: Do we need extra property. Why not just update `destinatonAddress` if it's empty?

Lykhoyda: It's true, it will not break any other logic, so we can keep only destinatonAddress

packages/widget/src/controllers/transfers/fungible-token-transfer.ts
```diff
@@ -114,8 +114,25 @@ export class FungibleTokenTransferController implements ReactiveController {
     void this.filterDestinationNetworksAndResources(network);
   };
 
+  getSenderDefaultDestinationAddress = (): void => {
```
mpetrunic, line 117:
```suggestion
  maybeSetSenderDefaultDestinationAddress = (): void => {
```

packages/widget/src/controllers/transfers/fungible-token-transfer.ts
```diff
@@ -114,8 +114,25 @@ export class FungibleTokenTransferController implements ReactiveController {
...
+
+    this.destinationAddress =
+      isSameNetwork || isSameType
+        ? this.walletContext.value?.evmWallet?.address || ''
+        : '';
+  };
+
   onDestinationNetworkSelected = (network: Domain | undefined): void => {
     this.destinationNetwork = network;
+    this.getSenderDefaultDestinationAddress();
```
mpetrunic, line 135:
```suggestion
    this.maybeSetSenderDefaultDestinationAddress();
```

mpetrunic (commented):
Last one I promise, feel free to choose different name though

packages/widget/src/controllers/transfers/fungible-token-transfer.ts
Lykhoyda: I'm not sure about "maybe" part, but as it's not returning the value the "setSenderDefaultDestinationAddress" looks sufficient for me. As empty value is also a value 😅

mpetrunic: I've added maybe because we aren't always setting it haha

mpetrunic (approved):

wainola (approved):

mpetrunic (approved):

Issue #93, opened 2024-02-14
Fungible Transfer logic from Substrate chain


mpetrunic:
## Implementation details

blocked by #74

LyonSsS:
<img width="563" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBNFhEQkE9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--61a15b214c685115eef7d360ffe3dfe1eed6a968/image.png" alt="image.png" />

PR #134, opened 2024-03-07
feat: substrate fungible transfer

wainola:
## Related Issue Or Context

Closes: #93 

## Types of changes

- [ ] Documentation

## Checklist:


packages/widget/src/controllers/transfers/substrate/execute.ts
```diff
@@ -0,0 +1,59 @@
...
+        this.host.requestUpdate();
+      }
+
+      if (isError) {
+        this.errorMessage = 'Transfer transaction reverted or rejected';
+        this.waitingTxExecution = false;
+        this.host.requestUpdate();
+      }
+
+      if (isFinalized) {
```
mpetrunic, line 52: this is not right. isFinalized means block if finalized. We don't need to handle this state

packages/widget/src/controllers/transfers/substrate/execute.ts
```diff
@@ -0,0 +1,59 @@
...
+    sender.address,
+    { signer: signer },
+    ({
+      isInBlock,
+      isFinalized,
+      blockNumber,
+      txIndex,
+      isError,
+      isCompleted
+    }) => {
```
mpetrunic, line 34: you should set `waitingUserConfirmation=false` when status isBroadcast=true.

packages/widget/src/controllers/wallet-manager/token-balance.ts
```diff
@@ -57,12 +67,21 @@ export class TokenBalanceController implements ReactiveController {
...
         );
         return;
       }
+    } else {
+      void this.suscribeSubstrateBalanceUpdate(resource);
+      this.timeout = setInterval(
+        this.suscribeSubstrateBalanceUpdate,
+        BALANCE_REFRESH_MS,
+        resource
+      ) as unknown as ReturnType<typeof setInterval>; //dubious
```
mpetrunic, line 83: why do you need this cast it's already of that type

packages/widget/src/controllers/wallet-manager/token-balance.ts
```diff
@@ -100,4 +119,25 @@ export class TokenBalanceController implements ReactiveController {
       this.host.requestUpdate();
     }.bind(this)();
   };
+
+  suscribeSubstrateBalanceUpdate = (resource: SubstrateResource): void => {
+    const { substrateProvider: apiPromise, accounts } = this.walletContext.value
+      ?.substrateWallet as SubstrateWallet;
+
+    const accountAddress = accounts![0].address;
```
mpetrunic, line 127: this should be signerAddress

packages/widget/src/controllers/wallet-manager/token-balance.ts
```diff
@@ -100,4 +119,25 @@ export class TokenBalanceController implements ReactiveController {
...
+
+    void async function (this: TokenBalanceController) {
+      const balance = await getAssetBalance(
```
mpetrunic, line 130: try/catch so you can update loadingBalance?

packages/widget/src/controllers/wallet-manager/token-balance.ts
```diff
@@ -100,4 +119,25 @@ export class TokenBalanceController implements ReactiveController {
...
+        apiPromise as ApiPromise,
+        (resource as unknown as { assetID: number }).assetID,
```
mpetrunic, line 132:
```suggestion
        (resource as SubstrateResource as { assetID: number }).assetID,
```
or whatever the name of type is

packages/widget/src/controllers/wallet-manager/token-balance.ts
```diff
@@ -100,4 +119,25 @@ export class TokenBalanceController implements ReactiveController {
...
+        accountAddress
+      );
+      this.loadingBalance = true;
+      this.host.requestUpdate();
+      this.loadingBalance = false;
+      this.balance = BigNumber.from(BigInt(balance.balance.toString()));
```
mpetrunic, line 138: what is the type of balance.balance?

packages/widget/src/controllers/wallet-manager/token-balance.ts
```diff
@@ -100,4 +119,25 @@ export class TokenBalanceController implements ReactiveController {
...
+      this.decimals = resource.decimals!;
```
mpetrunic, line 139: in which case is decimals undefined?

packages/widget/src/widget.ts
```diff
@@ -73,32 +82,53 @@ class SygmaProtocolWidget
...
-      <sygma-wallet-context-provider>
-        <section
-          class="widgetContainer ${this.isLoading ? 'noPointerEvents' : ''}"
+      <sygma-config-context-provider
+        .appMetadata=${this.appMetadata}
+        .theme=${this.theme}
+        .walletConnectOptions=${this.walletConnectOptions}
+      >
+        <sygma-wallet-context-provider
+          .substrateProvider=${'wss://rhala-node.phala.network/ws'}
```
mpetrunic, line 105: khm

packages/widget/src/context/wallet.ts
```diff
@@ -14,11 +15,12 @@ export interface EvmWallet {
...
 export interface SubstrateWallet {
-  signer: Signer;
-  signerAddress: string;
-  accounts: Account[];
+  signer?: Signer;
+  signerAddress?: string;
+  accounts?: Account[];
   unsubscribeSubstrateAccounts?: UnsubscribeFn;
   disconnect?: () => Promise<void>;
+  substrateProvider?: ApiPromise | string;
```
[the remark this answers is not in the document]
wainola: yeah makes sense, if the user can provide string to init the wsProvider, user can just provide ApiPromise.

packages/widget/src/context/wallet.ts
```diff
@@ -57,11 +59,28 @@ export class WalletContextProvider extends BaseComponent {
...
+  async connectedCallback(): Promise<void> {
     super.connectedCallback();
     if (this.evmWalllet) {
       this.walletContext.evmWallet = this.evmWalllet;
     }
+
+    let api: ApiPromise;
+
+    if (this.substrateProvider && typeof this.substrateProvider === 'string') {
+      const wssProvider = new WsProvider(this.substrateProvider);
```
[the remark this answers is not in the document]
wainola: I think with your last suggestion, we don't need this line anymore

packages/widget/src/controllers/transfers/substrate/execute.ts
```diff
@@ -0,0 +1,59 @@
...
+      if (isInBlock) {
+        this.waitingTxExecution = false;
+        this.host.requestUpdate();
+      }
+
+      if (isCompleted) {
```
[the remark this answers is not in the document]
wainola: you mean either isInBlock || isCompleted? or just use one of them?

packages/widget/src/controllers/transfers/substrate/execute.ts
wainola: so we handle the states update to display the status of the transfer on isInBlock or isCompleted?

packages/widget/src/controllers/wallet-manager/token-balance.ts
wainola: idk why the linter doesn't like that, considering that a couple of lines above is the same declaration 🤷 . I'm going to fix it

packages/widget/src/controllers/wallet-manager/token-balance.ts
wainola: could it be undefined since this relies on shared-config?

packages/widget/src/controllers/wallet-manager/token-balance.ts
wainola: linter is removing and replacing this for unknown. which is fine. This [PR](https://github.com/sygmaprotocol/sygma-sdk/pull/357) allows us to not type cast any special stuff

packages/widget/src/controllers/wallet-manager/token-balance.ts
wainola: u128

wainola: change name to tokenBalance to avoid having `balance.balance`.

packages/widget/src/controllers/wallet-manager/token-balance.ts
wainola: @mpetrunic I think we should also try/catch the evm ones, wdyt?

packages/widget/src/controllers/wallet-manager/token-balance.ts
mpetrunic: @MakMuftic Can decimals ever be undefined in shared config?

packages/widget/src/controllers/transfers/substrate/build.ts
```diff
@@ -0,0 +1,40 @@
...
+
+  const transfer = await substrateTransfer.createFungibleTransfer(
+    address,
+    this.destinationNetwork.chainId,
+    this.destinatonAddress,
+    this.selectedResource.resourceId,
+    String(this.resourceAmount)
+  );
+
+  const fee = await substrateTransfer.getFee(transfer);
```
saadjhk, line 34: can you see fee in a variable for UI to show?

packages/widget/src/context/wallet.ts
```diff
@@ -14,11 +16,12 @@ export interface EvmWallet {
 }
 
 export interface SubstrateWallet {
-  signer: Signer;
-  signerAddress: string;
-  accounts: Account[];
+  signer?: Signer;
+  signerAddress?: string;
+  accounts?: Account[];
```
saadjhk, line 21: why does `accounts` have to be `undefined`?

wainola: because of the way we need to provide `susbtrateProvider` as property that the widget is able to receive. Is either this or initialize by default the arguments of the substrate wallet when `ApiPromise` is created

packages/widget/src/controllers/transfers/substrate/build.ts
wainola: sure

packages/widget/src/controllers/transfers/substrate/execute.ts
wainola: updated

packages/widget/src/components/amount-selector/amount-selector.ts
```diff
@@ -51,9 +54,8 @@ export class AmountSelector extends BaseComponent {
 
   _onInputAmountChangeHandler = (event: Event): void => {
     const { value } = event.target as HTMLInputElement;
-    if (!this._validateAmount(value)) return;
-
     this.amount = Number.parseFloat(value);
+    if (!this._validateAmount(value)) return;
```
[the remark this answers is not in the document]
wainola: @mpetrunic here you are refering to L57 or L58? If we don't validate, are you suggesting to just pass the value to the lines below, provided that there is a `selectedResource`?

packages/widget/src/controllers/transfers/substrate/execute.ts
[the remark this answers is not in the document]
wainola: done

packages/widget/src/controllers/wallet-manager/token-balance.ts
wainola: Looks like I needed to reload vscode

packages/widget/src/controllers/wallet-manager/token-balance.ts
wainola: solved

packages/widget/src/components/amount-selector/amount-selector.ts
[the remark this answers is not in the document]
mpetrunic: It was 
˙˙`ts
    if (!this._validateAmount(value)) return;
    this.amount = Number.parseFloat(value);
```
and now is 
```ts
    this.amount = Number.parseFloat(value);
    if (!this._validateAmount(value)) return;
```
which means if user write some giberrish, `this.amount = Nan`

packages/widget/src/controllers/wallet-manager/token-balance.ts
```diff
@@ -100,4 +119,29 @@ export class TokenBalanceController implements ReactiveController {
...
+
+    void async function (this: TokenBalanceController) {
+      try {
+        this.loadingBalance = true;
+        const tokenBalance = await getAssetBalance(
+          apiPromise as ApiPromise,
+          resource.assetID as number,
+          signerAddress as string
+        );
+        this.host.requestUpdate();
```
mpetrunic, line 135: this should go before `await getAssetBalance˙

packages/widget/src/components/amount-selector/amount-selector.ts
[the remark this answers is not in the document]
wainola: this is very weird because I don't recall touching this 😆 .I see your point now

packages/widget/src/controllers/wallet-manager/token-balance.ts
wainola: totally true

packages/widget/src/context/wallet.ts
```diff
@@ -14,11 +16,12 @@ export interface EvmWallet {
...
   unsubscribeSubstrateAccounts?: UnsubscribeFn;
   disconnect?: () => Promise<void>;
+  substrateProvider?: ApiPromise;
```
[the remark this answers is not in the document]
wainola: Moved to it's own context provider

packages/widget/src/context/wallet.ts
```diff
@@ -31,6 +33,10 @@ export enum WalletContextKeys {
   SUBSTRATE_WALLET = 'substrateWallet'
 }
 
+export interface SubstrateProviderContext {
+  substrateProvider?: ApiPromise;
```
mpetrunic, line 37: Q: shouldn't widget user be able to provide multiple providers for each parachain. Like what happens if user switches source chain?

packages/widget/src/controllers/transfers/substrate/build.ts
```diff
@@ -0,0 +1,40 @@
...
+    !this.selectedResource ||
+    !this.destinatonAddress ||
+    !address
+  ) {
+    return;
+  }
+
+  const substrateTransfer = new SubstrateAssetTransfer();
+  await substrateTransfer.init(
+    this.substrateProviderContext.value?.substrateProvider as ApiPromise,
```
mpetrunic, line 22: Aren't you ignoring user selected source network here?

packages/widget/src/controllers/transfers/substrate/execute.ts
```diff
@@ -0,0 +1,47 @@
...
+        this.host.requestUpdate();
+      }
+
+      if (status.isBroadcast) {
+        this.waitingUserConfirmation = false;
+        this.host.requestUpdate();
+      }
+
+      if (isError) {
+        this.errorMessage = 'Transfer transaction reverted or rejected';
```
mpetrunic, line 41: Does reverted transaction ends up in block?

packages/widget/src/context/wallet.ts
wainola: I need to test this

packages/widget/src/controllers/transfers/substrate/build.ts
wainola: if I understand you correctly, we should init api promise based on the user selection, which makes total sense. Perhaps have the user to provide the collection of connections of multiple providers, user selects source and then upon matching them, we initialize the api promise object. Wdty?

packages/widget/src/context/wallet.ts
wainola: possible solution here: change type to `Array<{ domainId: number, provider: ApiPromise }>`, user sets the amount of providers he wants, as long as they are supported by us -which does not depend on us-, upon source network selection we match agains the `domainId`, init the provider and build the asset. Wdyt?

wainola: @mpetrunic I provided possible solution to this that also touches the comment below. Minor nitt here is that perhaps is better to use `chainId` instead of `domainId`

packages/widget/src/controllers/transfers/substrate/execute.ts
wainola: It does, I have changed this to deal with errors from extrinsic according to docs. I'm not entirely sure if we need to deal with other types of errors and this section is not being triggered at all when forcing this, like trying to deposit an amount that's not able to cover the fee.

packages/widget/src/context/wallet.ts
```diff
@@ -31,37 +33,70 @@ export enum WalletContextKeys {
   SUBSTRATE_WALLET = 'substrateWallet'
 }
 
+export interface SubstrateProviderContext {
+  substrateProviders?: Array<{ domainId: number; api: ApiPromise }>;
```
mpetrunic, line 37: wouldn't map make more sense so you can quickly fetch provider by domainId?

packages/widget/src/context/wallet.ts
```diff
@@ -31,37 +33,70 @@ export enum WalletContextKeys {
...
+  selectedProvider?: ApiPromise;
```
mpetrunic, line 38: Why do we need this? Context consumer can obtain provider corresponding to chainId needed in it's business logic. This context shouldn't care which network is selected

packages/widget/src/context/wallet.ts
```diff
@@ -31,37 +33,70 @@ export enum WalletContextKeys {
...
   private walletContext: WalletContext = {};
 
+  @provide({ context: substrateProviderContext })
+  private substrateProviderContext: SubstrateProviderContext = {};
+
   @property({ attribute: false, type: Object })
   evmWalllet?: EvmWallet;
 
+  @property({ attribute: false })
+  substrateProviders?: Array<{ domainId: number; api: ApiPromise }> = [];
```
mpetrunic, line 86: I don't think user should care about our internal domain id. He should just provide list of ApiPromise and we should from each fetch chainId and create map

mpetrunic (changes requested):

packages/widget/src/context/wallet.ts
```diff
@@ -31,6 +34,12 @@ export enum WalletContextKeys {
   SUBSTRATE_WALLET = 'substrateWallet'
 }
 
+export type ParachainProviders = Map<number, ApiPromise>;
```
mpetrunic, line 37: Maybe add comment explaining number is parachain id. Or create `type ParachainId=number;`

packages/widget/src/context/wallet.ts
```diff
@@ -54,14 +67,56 @@ export class WalletContextProvider extends BaseComponent {
...
+   */
+  async createProvidersMap(
+    providers: Array<ApiPromise>
+  ): Promise<ParachainProviders> {
+    const map: ParachainProviders = new Map();
+
+    for (const provider of providers) {
+      try {
+        // ! following pallet might not be available on all chains
+        // ? should we add polkadot type augmentations here?
```
mpetrunic, line 92: ideally yes

packages/widget/src/context/wallet.ts
```diff
@@ -54,14 +67,56 @@ export class WalletContextProvider extends BaseComponent {
...
+   * @returns {Promise<ParachainProviders>}
+   */
+  async createProvidersMap(
+    providers: Array<ApiPromise>
+  ): Promise<ParachainProviders> {
+    const map: ParachainProviders = new Map();
+
+    for (const provider of providers) {
+      try {
+        // ! following pallet might not be available on all chains
```
mpetrunic, line 91: I don't think we support parachains that doesn't have this pallet?

packages/widget/src/controllers/transfers/fungible-token-transfer.ts
```diff
@@ -133,9 +167,23 @@ export class FungibleTokenTransferController implements ReactiveController {
...
     this.sourceNetwork = network;
+    if (this.sourceNetwork.type === Network.SUBSTRATE) {
+      this.setSubstrateProvider(this.sourceNetwork);
+    }
     void this.filterDestinationNetworksAndResources(network);
   };
 
+  setSubstrateProvider = (sourceNetwork: Domain | undefined): void => {
+    const providers = this.substrateProviderContext.value?.substrateProviders;
+    if (!providers || !sourceNetwork?.chainId) {
```
mpetrunic, line 178:
```suggestion
    if (!providers || !sourceNetwork) {
```

packages/widget/src/controllers/transfers/fungible-token-transfer.ts
```diff
@@ -133,9 +167,23 @@ export class FungibleTokenTransferController implements ReactiveController {
...
+      return;
+    }
+
+    const provider = providers.get(sourceNetwork?.chainId);
```
mpetrunic, line 182:
```suggestion
    const provider = providers.get(sourceNetwork.chainId);
```

mpetrunic: Can you open issue for filtering supported source networks based on substrate providers we received from user?

packages/widget/src/controllers/transfers/fungible-token-transfer.ts
```diff
@@ -247,6 +296,11 @@ export class FungibleTokenTransferController implements ReactiveController {
         await getRoutes(this.env, sourceNetwork.chainId, 'fungible')
       );
     }
+
+    if (!this.substrateProvider && sourceNetwork.type === Network.SUBSTRATE) {
```
mpetrunic, line 300: why do we need this here?
Does it make more sense to not keep `this.substrateProvider` instead we replace it with getSubstrateProvider which returns provider based on selected source network?

packages/widget/src/context/wallet.ts
saadjhk: do we have a polkadot types package? i think we utilize that here @mpetrunic

mpetrunic: no idea, but we can install if needed

packages/widget/src/context/wallet.ts
```diff
@@ -54,14 +68,55 @@ export class WalletContextProvider extends BaseComponent {
...
+    const map: ParachainProviders = new Map();
+
+    for (const provider of providers) {
+      try {
+        // TODO: use polkadot type augmentation to remove "as 32"
+        const parachainId = await provider.query.parachainInfo.parachainId();
+        const _parachainId = (parachainId as u32).toNumber();
+        map.set(_parachainId, provider);
+      } catch (error) {
+        console.error(error);
```
mpetrunic, line 97:
```suggestion
        console.error("user provided substrate provider ignored", error);
```

saadjhk:
Ideally this should be merged once `parachainId` is available in `SubstrateConfig`

mpetrunic:
@saadjhk Latest version of sdk should have required changes

packages/widget/src/components/amount-selector/amount-selector.ts
```diff
@@ -1,22 +1,32 @@
...
 import { styles } from './styles';
 
 @customElement('sygma-resource-selector')
 export class AmountSelector extends BaseComponent {
   static styles = styles;
 
+  @property({
+    type: Object
+  })
+  sourceSubstrateProvider?: ApiPromise;
```
mpetrunic, line 28: I would remove this. Not really scalable if we start adding new chains. I would replace with sourceDomainId

packages/widget/src/components/amount-selector/amount-selector.ts
```diff
@@ -65,10 +74,25 @@ export class AmountSelector extends BaseComponent {
...
+  }
+
   _onResourceSelectedHandler = ({ value }: DropdownOption<Resource>): void => {
     this.selectedResource = value;
     this.amount = 0;
-    this.tokenBalanceController.startBalanceUpdates(value);
+
+    this.tokenBalanceController.startBalanceUpdates(
+      this.selectedResource,
+      this.sourceSubstrateProvider
```
mpetrunic, line 94: and pass sourceDomainId here.
Btw have you tried using widget? I'm getting unsupported resource for rococco-phala -> Sepolia because this is undefined. Should we hardcode some defaults?

packages/widget/src/controllers/transfers/substrate/execute.ts
```diff
@@ -0,0 +1,52 @@
...
+  const signer = this.walletContext.value?.substrateWallet?.signer;
+  if (
+    this.pendingTransferTransaction === undefined ||
+    destinationAddress == undefined ||
+    sender == undefined ||
+    this.sourceNetwork === undefined
+  )
+    return;
+
+  const provider = this.getSubstrateProvider(this.sourceNetwork?.chainId);
```
mpetrunic, line 21: shouldn't this be parachainId?

mpetrunic (changes requested):

packages/widget/src/components/amount-selector/amount-selector.ts
saadjhk: @mpetrunic so substrate providers need to be provided by the consumer as an attribute to the widget

mpetrunic: 1. We need better error message, indicating to the developer using widget what's going on
2. I would provide defaults at least in react example. But honestly I think we could also provide default rpc URLs for networks as well (user would override them if he provides them).

cc @MakMuftic @itsbobbyzzz168

saadjhk: @mpetrunic for keeping defaults widget will require urls of substrate chains to create ApiPromise objects

packages/widget/src/components/amount-selector/amount-selector.ts
saadjhk: @mpetrunic since amount-selector does not have access to `fungible-transfer-controller` state, having this as domain id will require passing down domain id -> parachain id mapping to use correct substrate provider, it is simpler to pass source parachain id IMO

packages/widget/src/components/amount-selector/amount-selector.ts
mpetrunic: I know, but we know those

saadjhk: @mpetrunic I don't think they are part of SDK or configuration

packages/widget/src/components/amount-selector/amount-selector.ts
mpetrunic: I'm trying to keep it generic for adding new chains. I don't think amount selector needs this it just passes it into TokenBalance controller which has access to substrate provider context

saadjhk: I understand but token balance controller does not have the context of domain <> parachain mapping. 

Since we are storing providers w.r.t parachain ID, providing a domain ID would require fetching parachain ID using domain ID and it wouldn't be possible as SDK is initialiazed in Fungible transfer controller. We can use `sourceDomainConfig` which may or may not include parachain id and is generic enough

wainola:
> conflicts (probably because of upstreamed changes from main)^^

solved

packages/widget/src/components/substrate-account-selector/substrate-account-selector.ts
```diff
@@ -76,7 +76,7 @@ export class SubstrateAccountSelector extends BaseComponent {
 
   render(): HTMLTemplateResult {
     const substrateWallet = this.wallets.substrateWallet;
-    const substrateAccount = substrateWallet?.accounts[0];
+    const substrateAccount = substrateWallet?.accounts![0];
```
mpetrunic, line 79:
```suggestion
    const substrateAccount = substrateWallet?.accounts?[0];
```
why exclamation?

packages/widget/src/constants.ts
```diff
@@ -1,5 +1,17 @@
+import { Environment } from '@buildwithsygma/sygma-sdk-core';
+
 export const DEFAULT_ETH_DECIMALS = 18;
 
 export const MAINNET_EXPLORER_URL = 'https://scan.buildwithsygma.com/transfer/';
 export const TESTNET_EXPLORER_URL =
   'https://scan.test.buildwithsygma.com/transfer/';
+
+export const SUBSTRATE_CHAIN_RPCS: { [env in Environment]: Array<string> } = {
```
mpetrunic, line 9: add comment (or event better adjust type) to only allow was connections

packages/widget/src/context/wallet.ts
```diff
@@ -54,14 +70,76 @@ export class WalletContextProvider extends BaseComponent {
...
   @property({ attribute: false, type: Object })
   evmWallet?: EvmWallet;
 
-  connectedCallback(): void {
+  @property({ attribute: false })
+  substrateProviders?: Array<ApiPromise> = [];
+
+  @property({ type: String }) environment?: Environment;
+
+  get defaultSubstrateProviders(): Array<ApiPromise> {
```
mpetrunic, line 84: maybe a nice solution would be instead of SUBSTRATE_CHAIN_RPCS url's to have parachainId -> ApiPromises direct mappings to avoid useless rpc requests 🤷🏻‍♂️

packages/widget/src/context/wallet.ts
```diff
@@ -54,14 +70,76 @@ export class WalletContextProvider extends BaseComponent {
...
+  async connectedCallback(): Promise<void> {
     super.connectedCallback();
     if (this.evmWallet) {
       this.walletContext.evmWallet = this.evmWallet;
     }
+
+    let substrateProviders;
+    if (this.substrateProviders) {
+      substrateProviders = await this.createProvidersMap(
+        this.substrateProviders
```
mpetrunic, line 131: why not combine with this.defaultSubstrateProviders?

mpetrunic (changes requested):
Please remove https://github.com/sygmaprotocol/sygma-widget/blob/462a5dbe919d1f6347018a5abb376a8216a5d8fd/packages/widget/src/controllers/transfers/fungible-token-transfer.ts#L194
and test the deployed flow before marking ready for review :(

packages/widget/src/constants.ts
saadjhk: you mean `ws` connections?

packages/widget/src/context/wallet.ts
saadjhk: @mpetrunic if we do that, we will make additional call even if substrate providers were provided by the third party. We dont need to have api promise objects if user have defined them.

packages/widget/src/constants.ts
mpetrunic:
```typescript
type WsUrl = `ws://${string}` | `wss://${string}`;
```
because later on, you are using this value as was connection

packages/widget/src/context/wallet.ts
saadjhk: `defaultSubstrateProviders` is only used when `substrateProviders` are not provided by the 3rd party

saadjhk: `defaultProviders` only makes calls when nothing is specified by user 🤔

mpetrunic: But why? Wouldn't it make more sense to always have defaults unless user overrides some or adds new ones?

saadjhk: @mpetrunic it does, and that's how it is right now. having ws urls hardcoded as strings is better than having default provider objects. if user overrides or provides theirs through property `substrateProviders` current logics pick them over the default rpc urls and no object is created.

What you are suggesting will always have provider objects regardless of user specifying them or not

packages/widget/src/context/wallet.ts
saadjhk: defaultSubstrateProviders should not be coupled with the specified providers logics imo. having a separate method is fine and `defaultSubstrateProviders` should always provide `ApiPromise` objects that are hardcoded or from our config

mpetrunic: I'm not sure I follow logic here.
Let's take for example, I am developer using this widget.
I have better endpoint for one parachain so I've put it into widget property. With current logic, only routes to parachain will work and other will not because default values are ignored. So I have to copy paste default values from widget.

More natural flow would be that I send one for which I have better endpoint and rest of the defaults stays the same.

Or do you envision this to work in some other way?

saadjhk: @mpetrunic okay i follow you now, so few questions
1. since defaultProviders are a must have due to the flow you suggested, is it fine to hardcode these on UI/widget?
2. Should we change the interface as well to accept a `Map` rather than array of providers?

packages/widget/src/context/wallet.ts
mpetrunic: Is this now more clear in regards to: https://github.com/sygmaprotocol/sygma-widget/pull/134#discussion_r1557630207 ? 😅

saadjhk: @mpetrunic yes it is! thank you very much

wainola:
> Please remove
> 
> https://github.com/sygmaprotocol/sygma-widget/blob/462a5dbe919d1f6347018a5abb376a8216a5d8fd/packages/widget/src/controllers/transfers/fungible-token-transfer.ts#L194
> 
> 
> and test the deployed flow before marking ready for review :(

filter remove by @saadjhk

mpetrunic (approved):
LFGG!

packages/widget/src/context/wallet.ts
```diff
@@ -31,6 +35,13 @@ export enum WalletContextKeys {
   SUBSTRATE_WALLET = 'substrateWallet'
 }
 
+export type ParachainId = number;
```
saadjhk, line 38: this is not needed as ParachainID is already provided by `@buildwithsygma`

saadjhk (approved):

packages/widget/src/context/wallet.ts
wainola: solved

saadjhk (approved):

mpetrunic (approved):

Issue #96, opened 2024-02-16
Retry and loading for getting bridge routes


mpetrunic:
## Implementation details
Without bridge routes, we cannot do much in the widget. I would  propose to retry infinitely and present full widget loading until we finally fetch routes (API has hardcoded values so only way this would fail if user doesn't have internet connection)

MakMuftic:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Sygma-Protocol-62f126a54b7d5a55203732f0/issues/sygmaprotocol/sygma-widget/96?planning-poker) with Zenhub @Lykhoyda @mpetrunic @wainola

mpetrunic:
closed via #136

PR #135, opened 2024-03-08
feat: Retry loading bridge routes

saadjhk:
## Description

Added retry sdk initialization

## Related Issue Or Context

https://github.com/sygmaprotocol/sygma-widget/issues/96

Closes: #96 

## How Has This Been Tested? Testing details.

- Tested locally

## Types of changes

Changes:
    

- [x] Added state in widget component to keep track of sdk initialization
- [x] Added event to be fired when sdk is initialized
- [x] Retry sdk init method

### DIFF
--- a/packages/widget/src/controllers/transfers/fungible-token-transfer.ts
+++ b/packages/widget/src/controllers/transfers/fungible-token-transfer.ts
@@ -11,6 +11,7 @@ import { BigNumber } from 'ethers';
 import type { ReactiveController, ReactiveElement } from 'lit';
 import { walletContext } from '../../context/wallet';
 import { MAINNET_EXPLORER_URL, TESTNET_EXPLORER_URL } from '../../constants';
+import { SdkInitializedEvent } from '../../interfaces';
 import { buildEvmFungibleTransactions, executeNextEvmTransaction } from './evm';
 
 export enum FungibleTransferState {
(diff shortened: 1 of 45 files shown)


packages/widget/src/controllers/transfers/fungible-token-transfer.ts
```diff
@@ -80,10 +81,29 @@ export class FungibleTokenTransferController implements ReactiveController {
...
+   * @returns {void}
+   */
+  async retryInitSdk(): Promise<void> {
+    try {
+      await this.config.init(1, this.env);
+      this.host.dispatchEvent(
+        new SdkInitializedEvent({ hasInitialized: true })
+      );
+    } catch (error) {
+      // Add a sleep to avoid thousands of calls?
```
mpetrunic, line 97: 100%. Probably increase timeout which every call

packages/widget/src/widget.ts
```diff
@@ -65,9 +66,13 @@ class SygmaProtocolWidget
 
   @property({ type: Object }) appMetadata?: AppMetadata;
 
+  // ? where is this used?
```
mpetrunic, line 69: on L131, maybe use this instead of new sdkInitialized state? Or delete isLoading haha

mpetrunic (changes requested):
Please cherrypick your changes onto new branch/pr originating from dev branch as you are now trying to merge hotfixes from main

Issue #123, opened 2024-02-27
Max button is disabled after allow spending is approved once


LyonSsS:
**Context**:  After the user allows spending once, the UI disable the max button. 
![Max_not_work](https://github.com/sygmaprotocol/sygma-widget/assets/30255634/b1e87d39-f4ae-4f01-9803-e78eb0548431)

MakMuftic:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Sygma-Protocol-62f126a54b7d5a55203732f0/issues/sygmaprotocol/sygma-widget/123?planning-poker) with Zenhub @Lykhoyda @mpetrunic @wainola

Lykhoyda:
@mpetrunic should be solved by this PR https://github.com/sygmaprotocol/sygma-widget/pull/137

mpetrunic:
@Lykhoyda Please just add that into description of your PR :)

LyonSsS:
Retested on https://github.com/sygmaprotocol/sygma-widget/pull/137 but it is still an issue 
![image](https://github.com/sygmaprotocol/sygma-widget/assets/30255634/ab7e0be2-c84c-459a-8bcf-4f92156c73a8)
![image](https://github.com/sygmaprotocol/sygma-widget/assets/30255634/11a20f46-e156-49e0-b250-8b6e13571e8c)

Lykhoyda:
Fixed in PR https://github.com/sygmaprotocol/sygma-widget/pull/137

LyonSsS:
Retested on Core release branch and works fine 
![image](https://github.com/sygmaprotocol/sygma-widget/assets/30255634/2b33b9c0-372d-4bdd-bdec-fe4e93993a49)

PR #137, opened 2024-03-11
fix: max button rounds value

### LINKED ISSUE #124: MAX amount rounds value to 2 decimals which fails at allow spending or transfer, if value is below rounding
Severity: Medium | Priority: High
**Context**:  When using MAX button, allow spending and transfer are affected if the value is rounded >= actual balance. 

Actual balance on chain: 
image

Allow spending behavior: 
MAX_AllowSpending

Transfer Max ( after allow spending was done): 
MAX_Transfer

Lykhoyda:
Closes #124 #123


wainola:
@mpetrunic @Lykhoyda how many decimals do we want to display if user clicks on max? currently we display 4. Do we want to adjust that?

packages/widget/src/components/amount-selector/amount-selector.ts
```diff
@@ -45,26 +48,31 @@ export class AmountSelector extends BaseComponent {
 
   tokenBalanceController = new TokenBalanceController(this);
 
+  _handleOnResourceSelectedCallback = (): void => {
+    if (this.selectedResource) {
+      this.onResourceSelected(
+        this.selectedResource,
+        utils.parseUnits(
+          this.amount.toString(),
+          this.selectedResource.decimals ?? DEFAULT_ETH_DECIMALS
```
wainola, line 57: I think is almost impossible for a resource to have undefined decimals. Because this info either is comming from `shared-config` or is going to come from another source of truth, where this is going to be hardcoded and defined. I'm not sure here if we need the `DEFAULT_ETH_DECIMALS` here 🤔

packages/widget/src/utils/token.ts
```diff
@@ -7,3 +7,21 @@ export function tokenBalanceToNumber(
...
 }
+
+export function truncateBigNumberToString(
+  value: BigNumber,
+  decimals: number,
+  displayDecimals: number = 4
+): string {
+  const valueStr = utils.formatUnits(value, decimals);
+
+  const parts = valueStr.split('.');
```
wainola, line 18: can we do here something like `Number(valueStr).toFixed(displayDecimals)`? You avoid having to split and slice stuff here. If `displayDecimals` is 0, we can just return the number

wainola (changes requested):

packages/widget/src/components/amount-selector/amount-selector.ts
```diff
@@ -45,26 +48,31 @@ export class AmountSelector extends BaseComponent {
 
   tokenBalanceController = new TokenBalanceController(this);
 
+  _handleOnResourceSelectedCallback = (): void => {
```
mpetrunic, line 51: Maybe better to move that into existing resourceSelected callback: https://github.com/sygmaprotocol/sygma-widget/pull/137/files#diff-27d2b4e4c285a859d1c78d56d9586464abcf08c45067c54bbb0d17a322fddfb2R89 ?

mpetrunic: Also why is this change in this PR?

mpetrunic (changes requested):
Way too complicated. Can you redo PR by using following code:
```ts
Math.floor(Number.parseFloat(utils.formatUnits(BigNumber.from("28999999998000000000"), 18))*10000) / 10000
```

packages/widget/src/utils/token.ts
Lykhoyda: @wainola This was an initial issue toFixed will round the balance resulting in having more balance than the user has. [Reference about toFixed](https://medium.com/@stheodorejohn/javascripts-tofixed-precision-in-your-hands-89ff981f50a9#:~:text=Rounding%20Behavior%3A%0AtoFixed()%20uses%20standard%20rounding%20behavior%2C%20so%20values%20are%20rounded%20to%20the%20nearest%20decimal%20place.%20Be%20aware%20that%20this%20may%20not%20always%20match%20your%20specific%20rounding%20requirements.)

packages/widget/src/components/amount-selector/amount-selector.ts
sztok7: i think the logic is extracted because it's used in 2 places (i usually do this when i see it's used >=3 times) 🤷

packages/widget/src/components/transfer/fungible/transfer-status/transfer-status.ts
```diff
@@ -44,7 +45,7 @@ export class TransferStatus extends BaseComponent {
         </span>
       </div>
       <div class="tokenInfo">
-        ${this.formatAmount(this.amount ?? BigNumber.from(0))}
+        ${this.formatAmount(this.amount ?? ethers.constants.Zero)}
```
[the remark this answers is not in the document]
Lykhoyda: We can remove it, as well as some code above which is checking for an empty string like 
`' ${this.renderNetworkIcon(this.destinationNetworkName ?? '')}'`

packages/widget/src/utils/token.ts
wainola: oh I see, thanks for the clarification

packages/widget/src/components/amount-selector/amount-selector.ts
Lykhoyda: Maybe it's true, but the interface from sygma-sdk-core tells that it could be undefined, so we keep the default value

packages/widget/src/components/amount-selector/amount-selector.ts
Lykhoyda: Exactly, I created a method as I used it twice. The problem with the max button is when we set the value this.amount the "@change"  _onInputAmountChangeHandler will not be triggered to call this.onResourceSelected inside the method. So I call it manually inside the _useMaxBalance

Lykhoyda:
> Way too complicated. Can you redo PR by using following code:
> 
> ```ts
> Math.floor(Number.parseFloat(utils.formatUnits(BigNumber.from("28999999998000000000"), 18))*10000) / 10000
> ```

I agree it's more complicated, but it will work with bigger numbers. Might be overkill for this scenario as we have max 18 decimals. So I replace the code with yours

packages/widget/src/utils/token.ts
```diff
@@ -3,25 +3,19 @@ import { utils } from 'ethers';
...
+    return tokenBalanceToNumberWithDecimals(value);
+  }
 
-  if (displayDecimals === 0) return integerPart;
+  return value;
+}
 
-  return `${integerPart},${decimalPart}`;
+// This function is used to round the token balance to 4 decimal places
+function tokenBalanceToNumberWithDecimals(value: number): number {
```
mpetrunic, line 19: maybe make more generic and allow passing number of decimals?

packages/widget/src/utils/token.ts
```diff
@@ -3,25 +3,19 @@ import { utils } from 'ethers';
...
-  value: BigNumber,
   decimals: number,
-  displayDecimals: number = 4
-): string {
-  const valueStr = utils.formatUnits(value, decimals);
-
-  const parts = valueStr.split('.');
-  const integerPart = parts[0];
-  let decimalPart = parts.length > 1 ? parts[1] : '';
+  withFormattedDecimals = false
```
mpetrunic, line 7: maybe instead let us pass number of decimals we would like to see

mpetrunic (approved):

wainola (approved):

mpetrunic (dismissed):

LyonSsS (approved):
Booth QA tickets related to this passed. This is approved

packages/widget/src/components/amount-selector/amount-selector.ts
```diff
@@ -68,6 +61,23 @@ export class AmountSelector extends BaseComponent {
...
+    const amount = utils.parseUnits(
+      value,
+      this.tokenBalanceController.decimals
+    );
+    this.amount = utils.formatUnits(
+      amount,
+      this.tokenBalanceController.decimals
+    );
+
+    if (!this._validateAmount(value)) return;
```
mpetrunic, line 75: this has to go before parsing, not sure how it keeps ending up after which causes errors if user inputs not a number

mpetrunic (changes requested):

packages/widget/src/components/amount-selector/amount-selector.ts
Lykhoyda: I double-checked that, and I think the order is right if it's a numeric value that doesn't meet the criteria f.e more than the balance we still need to set it to "this.amount" to keep the input in sync. Regarding your concern, if the user inputs some non-numeric value I wrap the code in a try-catch block if the parse fails.

mpetrunic (approved):

sztok7 (approved):
LGTM

PR #144, opened 2024-03-14
feat: Transfer UI page 

mj52951:
Created a transfer UI page by refactoring `react-example`. 

Centered the Widget React component and created a footer with links to other Sygma pages.


MakMuftic (changes requested):
Hey @mj52951 can we make a few small changes:
- Make footer and faucet text stick to the bottom of the page
- Have some margin between the top of the widget and the top of the page
- Change the icon of the docs so it is not the same as the website

examples/react-widget-app/src/App.css
```diff
@@ -0,0 +1,54 @@
+html {
+    background-color: #F5F0E9;
+    font-family: -apple-system, BlinkMacSystemFont, 'Segoe UI', 'Roboto', 'Oxygen',
```
wainola, line 3: I think we don't need this, since there is a font definition on widget... if we need this for the other divs you created, better to use the same font that we are using on the widget

examples/react-widget-app/src/App.css
```diff
@@ -0,0 +1,54 @@
...
+}
+
+.centered {
+    display: flex;
+    justify-content: center;
+    align-items: center;
+}
+
+.text-container {
+    font-size: 20px;
```
wainola, line 15: transfomr all the px to rem pls 😄

wainola (changes requested):

wainola:
@MakMuftic would it make sense to have footer as header instead?

wainola:
@mj52951 are the comments really resolved? if not could you provide a comment to update?

MakMuftic (changes requested):
As discussed in person @mj52951, let us move this additional information to the sidebar menu. We can use this in the future to add the ability to configure widget properties live on this page.

mpetrunic (approved):

wainola (approved):

Issue #55, opened 2024-01-15
Display fee


FSM1:
## Story
As a bridge user, I want to know which fees I will be paying to bridge tokens.

## Background
The fee, as well as the token the fee is charged in should be displayed.
 
## Details
Screenshot is provisional
![image](https://github.com/sygmaprotocol/sygma-widget/assets/12774278/7ab32293-199d-409c-bd75-b04b356e94b2)

## Testing details
Test both Fixed and Percentage fee

## Acceptance criteria
Fee and FeeToken are displayed correctly.

MakMuftic:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Sygma-Protocol-62f126a54b7d5a55203732f0/issues/sygmaprotocol/sygma-widget/55?planning-poker) with Zenhub @Lykhoyda @mpetrunic @wainola

PR #145, opened 2024-03-15
feat: Display Fee

saadjhk:
This PR adds UI in widget to display fee that will be charged on `FungibleTransfer`

## Description

- [X] Added Transfer detail component
- [X] Modified `buildEvmFungibleTransactions` to store fee in component state
- [X] Fixed button visibility issue

## Related Issue Or Context

https://github.com/sygmaprotocol/sygma-widget/issues/55

Closes: #55 

## Types of changes

- [X] Component Added with styles
- [X] Fixed button visibility issue
- [X] Changes added to `buildTransactions` and `buildEvmTransactions` methods

## Checklist:
- [ ] Add `sygma-fungible-transfer-detail` tests

### DIFF
--- a/packages/widget/src/components/transfer/fungible/transfer-detail/transfer-detail.ts
+++ b/packages/widget/src/components/transfer/fungible/transfer-detail/transfer-detail.ts
@@ -27,37 +27,39 @@ export class FungibleTransferDetail extends BaseComponent {
   @property({ type: Object })
   sourceDomainConfig?: BaseConfig<Network>;
 
-  displayFee(): string {
-    if (!this.fee) return '';
-    let balance = '';
+  getFeeParams(type: FeeHandlerType): { decimals: string; symbol: string } {
+    let decimals = '';
     let symbol = '';
-    let decimals: number | undefined = undefined;
-    const { type, fee } = this.fee;
 
-    if (type === FeeHandlerType.PERCENTAGE) {
-      if (this.selectedResource) {
-        if (this.selectedResource.symbol) {
-          symbol = this.selectedResource.symbol;
+    switch (type) {
+      case FeeHandlerType.BASIC:
+        if (this.sourceDomainConfig) {
+          decimals = this.sourceDomainConfig.nativeTokenDecimals.toString();
+          symbol = this.sourceDomainConfig.nativeTokenSymbol.toUpperCase();
         }
-
-        if (this.selectedResource.decimals) {
-          decimals = this.selectedResource.decimals;
+        return { decimals, symbol };
+      case FeeHandlerType.DYNAMIC:
+        if (this.selectedResource) {
+          symbol = this.selectedResource.symbol ?? '';
+          decimals = this.selectedResource.decimals?.toString() ?? '';
         }
-      }
+        return { decimals, symbol };
+      default:
+        return { decimals, symbol };
+    }
+  }
+
+  getFee(): string {
+    if (!this.fee) return '';
+    const { symbol, decimals } = this.getFeeParams(this.fee.type);
+    const { fee } = this.fee;
+    let _fee = '';
+
+    if (decimals.length > 0) {
+      _fee = tokenBalanceToNumber(fee, Number(decimals)).toFixed(4);
     }
 
-    if (type === FeeHandlerType.BASIC) {
-      if (this.sourceDomainConfig) {
-        symbol = this.sourceDomainConfig.nativeTokenSymbol.toUpperCase();
-        decimals = Number(this.sourceDomainConfig.nativeTokenDecimals);
-      }
-    }
-
-    if (decimals) {
-      balance = tokenBalanceToNumber(fee, decimals).toFixed(4);
-    }
-
-    return `${balance} ${symbol}`;
+    return `${_fee} ${symbol}`;
   }
 
   render(): HTMLTemplateResult {
@@ -67,10 +69,8 @@ export class FungibleTransferDetail extends BaseComponent {
           this.fee !== undefined,
           () =>
             html`<div class="transferDetailContainer">
-              <div class="transferDetailContainer-label">Fee</div>
-              <div class="transferDetailContainer-value">
-                ${this.displayFee()}
-              </div>
+              <div class="transferDetailContainerLabel">Fee</div>
+              <div class="transferDetailContainerValue">${this.getFee()}</div>
             </div>`
         )}
       </section>
--- a/packages/widget/src/components/transfer/fungible/transfer-detail/transfer-detail.ts
+++ b/packages/widget/src/components/transfer/fungible/transfer-detail/transfer-detail.ts
@@ -38,7 +38,7 @@ export class FungibleTransferDetail extends BaseComponent {
           symbol = this.sourceDomainConfig.nativeTokenSymbol.toUpperCase();
         }
         return { decimals, symbol };
-      case FeeHandlerType.DYNAMIC:
+      case FeeHandlerType.PERCENTAGE:
         if (this.selectedResource) {
           symbol = this.selectedResource.symbol ?? '';
           decimals = this.selectedResource.decimals ?? undefined;
@@ -69,7 +69,7 @@ export class FungibleTransferDetail extends BaseComponent {
           this.fee !== undefined,
           () =>
             html`<div class="transferDetailContainer">
-              <div class="transferDetailContainerLabel">Fee</div>
+              <div class="transferDetailContainerLabel">Bridge Fee</div>
               <div class="transferDetailContainerValue">${this.getFee()}</div>
             </div>`
         )}
--- a/packages/widget/src/components/transfer/fungible/transfer-detail/transfer-detail.ts
+++ b/packages/widget/src/components/transfer/fungible/transfer-detail/transfer-detail.ts
@@ -1,0 +1,70 @@
+import {
+  FeeHandlerType,
+  type Domain,
+  type EvmFee,
+  type Resource
+} from '@buildwithsygma/sygma-sdk-core';
+import '../../../common/buttons/button';
+import type { HTMLTemplateResult } from 'lit';
+import { html } from 'lit';
+import { when } from 'lit/directives/when.js';
+import { customElement, property } from 'lit/decorators.js';
+import { tokenBalanceToNumber } from '../../../../utils/token';
+import { BaseComponent } from '../../../common/base-component';
+import { styles } from './styles';
+
+@customElement('sygma-fungible-transfer-detail')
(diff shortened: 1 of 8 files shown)


packages/widget/src/components/transfer/fungible/transfer-detail/transfer-detail.ts
```diff
@@ -0,0 +1,87 @@
...
+          decimals = Number(domainConfig.nativeTokenDecimals);
+        }
+      }
+    }
+
+    if (decimals) {
+      balance = tokenBalanceToNumber(this.fee.fee, decimals).toFixed(4);
+    }
+
+    // ? dynamic
```
mpetrunic, line 65: doesn't exist right now. It was deprectaed

packages/widget/src/components/transfer/fungible/transfer-detail/transfer-detail.ts
```diff
@@ -0,0 +1,87 @@
...
+    let decimals: number | undefined = undefined;
+
+    if (this.fee.type === FeeHandlerType.PERCENTAGE) {
+      if (this.selectedResource) {
+        if (this.selectedResource.symbol) {
+          symbol = this.selectedResource.symbol;
+        }
+
+        if (this.selectedResource.decimals) {
+          decimals = this.selectedResource?.decimals;
```
mpetrunic, line 46:
```suggestion
          decimals = this.selectedResource.decimals;
```
Are you sure you need optional chaining here?

packages/widget/src/components/transfer/fungible/transfer-detail/transfer-detail.ts
```diff
@@ -0,0 +1,87 @@
...
+  fee?: EvmFee;
+
+  @property({ type: Object })
+  selectedResource?: Resource;
+
+  @property({ type: Object })
+  sourceNetwork?: Domain;
+
+  @property({ type: Object })
+  config?: Config;
```
mpetrunic, line 31: maybe you can replace config and sourceNetwork properties with DomainConfig property?

packages/widget/src/controllers/transfers/fungible-token-transfer.ts
```diff
@@ -62,6 +68,10 @@ export class FungibleTokenTransferController implements ReactiveController {
   host: ReactiveElement;
   walletContext: ContextConsumer<typeof walletContext, ReactiveElement>;
 
+  get __config(): Config {
```
mpetrunic, line 71: maybe change this for `getSourceNetworkConfig` which will return DomainConfig for selected network?

packages/widget/src/components/transfer/fungible/transfer-detail/transfer-detail.ts
saadjhk: @mpetrunic its not needed, ill remove

packages/widget/src/components/transfer/fungible/transfer-detail/transfer-detail.ts
```diff
@@ -0,0 +1,85 @@
...
+  sourceDomainConfig?: BaseConfig<Network>;
+
+  displayFee(): string {
+    if (!this.fee) return '';
+    let balance = '';
+    let symbol = '';
+    let decimals: number | undefined = undefined;
+    const { type, fee } = this.fee;
+
+    if (type === FeeHandlerType.PERCENTAGE) {
```
Lykhoyda, line 37: In the case of "type" we can probably introduce a switch case to improve the readability. Plus if we would use another possible type from FeeHandlerType, the code will be "future ready".

packages/widget/src/components/transfer/fungible/transfer-detail/transfer-detail.ts
```diff
@@ -0,0 +1,85 @@
...
+      if (this.selectedResource) {
+        if (this.selectedResource.symbol) {
+          symbol = this.selectedResource.symbol;
+        }
+
+        if (this.selectedResource.decimals) {
+          decimals = this.selectedResource.decimals;
+        }
```
Lykhoyda, line 45: We can get rid of nested "if's" using optional chaining

packages/widget/src/components/transfer/fungible/transfer-detail/transfer-detail.ts
```diff
@@ -0,0 +1,85 @@
...
+
+    if (type === FeeHandlerType.BASIC) {
+      if (this.sourceDomainConfig) {
+        symbol = this.sourceDomainConfig.nativeTokenSymbol.toUpperCase();
+        decimals = Number(this.sourceDomainConfig.nativeTokenDecimals);
+      }
+    }
+
+    if (decimals) {
+      balance = tokenBalanceToNumber(fee, decimals).toFixed(4);
```
Lykhoyda, line 57: Pull the latest code, our method will support desired amount of decimals as a parameter. So the "toFixed" will be redundant. Also please pay attention that the method works with the "BigNumber"

packages/widget/src/components/transfer/fungible/transfer-detail/transfer-detail.ts
```diff
@@ -0,0 +1,85 @@
...
+    const { type, fee } = this.fee;
+
+    if (type === FeeHandlerType.PERCENTAGE) {
+      if (this.selectedResource) {
+        if (this.selectedResource.symbol) {
+          symbol = this.selectedResource.symbol;
+        }
+
+        if (this.selectedResource.decimals) {
+          decimals = this.selectedResource.decimals;
```
Lykhoyda, line 44: It's probably better to separate the getter of fee details and the display, f.e:
```
 getFeeDetails(): { symbol: string; decimals?: number } {
    switch (this.fee?.type) {
      case FeeHandlerType.PERCENTAGE:
        return {symbol: '', decimal: ''}; // logic is missing
      case FeeHandlerType.BASIC:
        return {symbol: '', decimal: ''}; // logic is missing
      default:
        return { symbol: '', decimals: undefined };
    }
  }
```

packages/widget/src/components/transfer/fungible/transfer-detail/transfer-detail.ts
```diff
@@ -0,0 +1,85 @@
...
+
+  render(): HTMLTemplateResult {
+    return html`
+      <section class="transferDetail">
+        ${when(
+          this.fee !== undefined,
+          () =>
+            html`<div class="transferDetailContainer">
+              <div class="transferDetailContainer-label">Fee</div>
+              <div class="transferDetailContainer-value">
```
Lykhoyda, line 71: Due to issues with the parsing of "-" it's better to keep camelCase classes similar to other code. We discussed switching to BEM, but I faced issues with the classNames when using "part pseudo element". For consistency's sake, I recommend using a camel case for now.  https://developer.mozilla.org/en-US/docs/Web/CSS/::part

packages/widget/src/controllers/transfers/fungible-token-transfer.ts
```diff
@@ -47,6 +54,7 @@ export class FungibleTokenTransferController implements ReactiveController {
   public supportedSourceNetworks: Domain[] = [];
   public supportedDestinationNetworks: Domain[] = [];
   public supportedResources: Resource[] = [];
+  public fee?: EvmFee;
```
Lykhoyda, line 57: Can the fee be undefined? My understanding that it's always should be defined, in case the fee is not applicable it will be zero instead. @mpetrunic can you help me here to understand what are the other factors that would influence whether the fee exists or not?

mpetrunic: You need to set all the fields to be able to calculate and set fee in this field. Fee depends on route/resource chosen and amount

packages/widget/tests/unit/components/transfer-detail/transfer-detail.test.ts
```diff
@@ -0,0 +1,68 @@
...
+        .selectedResource=${mockedResource}
+        .sourceDomainConfig=${mockedSourceDomainConfig}
+      ></sygma-fungible-transfer-detail>
+    `);
+
+    const transferDetail = el.shadowRoot!.querySelector(
+      '.transferDetail'
+    ) as HTMLElement;
+
+    assert.isNotNull(transferDetail);
```
Lykhoyda, line 66: can we check explicitly the calculated fee inside "transferDetailContainer-value"?

Lykhoyda:
@saadjhk It will be nice also to have the font sizes as per design here: (https://www.figma.com/file/T3A0opRIH07hSqUaESKgFV/Sygma-Widget-UI?type=design&node-id=285-5497&mode=dev)
And the label "Bridge fee" is more informative than just "fee". Can you please align with the design?
Screenshot 2024-03-19 at 13 37 45

packages/widget/src/components/transfer/fungible/transfer-detail/transfer-detail.ts
Lykhoyda: Update: the latest update is in the master branch plus all the fixes. We need to sync them

Lykhoyda (changes requested):

packages/widget/src/components/transfer/fungible/transfer-detail/transfer-detail.ts
saadjhk: @Lykhoyda can we merge those changes in dev?

Lykhoyda:
⚠️ To be merged before this PR https://github.com/sygmaprotocol/sygma-widget/pull/149

mpetrunic (approved):

Lykhoyda:
@saadjhk it gets stuck with the error "Failed getting fee", so I couldn't test the entire UI. Is it expected for now, can you please take a look?


Screenshot 2024-03-26 at 15 48 54

saadjhk:
@

> @saadjhk it gets stuck with the error "Failed getting fee", so I couldn't test the entire UI. Is it expected for now, can you please take a look?
> Screenshot 2024-03-26 at 15 48 54

seems like a sdk level error, there is no error handling on fetching fee so routes with no fee configuration must throw this error. However, I'm not sure if that should be handled in this PR

Lykhoyda (approved):

mpetrunic (approved):

mpetrunic:
> @
> 
> > @saadjhk it gets stuck with the error "Failed getting fee", so I couldn't test the entire UI. Is it expected for now, can you please take a look?
> > Screenshot 2024-03-26 at 15 48 54
> 
> seems like a sdk level error, there is no error handling on fetching fee so routes with no fee configuration must throw this error. However, I'm not sure if that should be handled in this PR

Please open issue on sdk repo!

Issue #133, opened 2024-03-06
Source network gets reset on starting new transfer


mpetrunic:
## Expected Behavior
If wallet is still connected we shouldn't reset selected source network.

## Current Behavior
We reset selected source network which ends us in unsupported state where wallet is connected before source network is selected

## Steps to Reproduce (for bugs)

1.
2.
3.
4.

## Version

Sygma commit (or docker tag):
sygma-solidity version:
sygma-substrate version:
Go version:

LyonSsS:
From my point of view, since the specifications and also Figma design flow, do not specify this behavior, I do consider that a "starting a new transfer" = "initial transfer". Because of this, I do consider the behavior as normal. 
I would be glad with booth ways - restart OR keep the user historical data as it was before ( Source + Dest and TOKEN type selected, but nothing more).
I think this can be of course answered by product.  CC: @itsbobbyzzz168

mpetrunic:
I think question is do we disconnect wallet as well or do we leave source network selected and remove everything else. As those are only two valid states for widget

LyonSsS:
Teste on PR  link
![Restart_Transatcion](https://github.com/sygmaprotocol/sygma-widget/assets/30255634/1cba026b-f450-4b27-a73c-b7fd8117a122)

PR #152, opened 2024-03-21
fix: clear wallet context on disconnect

### LINKED ISSUE #113: Reset context when disconnecting/lock wallet
Severity: low | Priority: low

**Note**: After connecting Metamask wallet, and setting everything ready for a transfer, if the user disconnects the wallet via disconnect or lock ( metamask option), the context is not reseted. 

image

Lykhoyda:
Closes #113

The original issue, mentions NOT clear the selected and destination network. But we fetch the resource data and other data on selecting the destination network, so to avoid unclear behavior for the user to click again the destination network I clear all the data. If it's important to have the behavior to keep the state of networks, it would be better to keep at least a selected network and a clear destination network to avoid potential bugs with the data being not updated on component levels. IMO clearing the entire state is the most bulletproof way to avoid previous selected data


packages/widget/src/components/resource-amount-selector/resource-amount-selector.ts
```diff
@@ -43,6 +41,16 @@ export class AmountSelector extends BaseComponent {
   @state() validationMessage: string | null = null;
   @state() amount: string = '0';
 
+  constructor() {
+    super();
+    document.addEventListener('wallet-disconnected', () => {
```
mpetrunic, line 46: instead of doing this everywhere, can you move this logic into transferController?

packages/widget/src/controllers/wallet-manager/manager.ts
```diff
@@ -61,6 +61,9 @@ export class WalletController implements ReactiveController {
   disconnectWallet = (): void => {
     this.disconnectEvmWallet();
     this.disconnectSubstrateWallet();
+    this.host.dispatchEvent(
```
mpetrunic, line 64: Don't think you need this. TransferController is subscribed to walletContext changes https://github.com/sygmaprotocol/sygma-widget/blob/dev/packages/widget/src/controllers/transfers/fungible-token-transfer.ts#L71

so you can check is wallet is disconnected on every change

mpetrunic (changes requested):

packages/widget/src/components/resource-amount-selector/resource-amount-selector.ts
Lykhoyda: good idea! I tried this approach but from the first try it didn't go well :) Please check the updated code

mpetrunic:
This doesn't resolve #133. Please open new PR to address the issue.Right now if I start new transfer, source network is resetted while wallet remains connected

mpetrunic (approved):

Issue #117, opened 2024-02-26
Switching to a different TOKEN mapping for "From - TO" relationship is not updated in UI if amount is inserted


LyonSsS:
Severity: Low | Priority: Medium

**Context**: When switching to another "From - To" mapping, that needs to update the available compatible ERC20 token vs previous selection, UI is not updated. 

**STR**:
1. Login with metmask and select from as Sepolia.
2. Select to as Mumbai and select drop down to ERC20RLTest tokens.
3. After balance is updated, insert an amount <= MAX in the "Amount to transfer" field. 
4. Select TO destination to Rhala, which has the only mapping to ERC20 token sygUSDC. 

**Exp**: UI to update the correct erc20 mapping.
**ACT**: Nothing happened. The same ERC20RLTest token remain. 

![image](https://github.com/sygmaprotocol/sygma-widget/assets/30255634/aff1c17b-d907-4fdd-ac26-7e26252f19dd)

MakMuftic:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Sygma-Protocol-62f126a54b7d5a55203732f0/issues/sygmaprotocol/sygma-widget/117?planning-poker) with Zenhub @Lykhoyda @mpetrunic @wainola

saadjhk:
@LyonSsS I tested this functionality by pulling latest commits from `main` and it seems to be fixed, can you please validate the functionality once?

LyonSsS:
I did retested it on latest build from https://github.com/sygmaprotocol/sygma-widget/pull/60 and I could reproduce it. 
![image](https://github.com/sygmaprotocol/sygma-widget/assets/30255634/045b1b2b-032c-415f-8702-af31ac6fede0)

LyonSsS:
Retested on CoreReleaseBranch - this now works fine

PR #155, opened 2024-03-25
fix: Switching to a different TOKEN mapping for "From - TO" relationship is not updated in UI if amount is inserted

saadjhk:
- Added network type to address input component
- remove selected resource on destination network change

## Description

Resource selection is not updated on UI when a destination network is changed.

## Related Issue Or Context

https://github.com/sygmaprotocol/sygma-widget/issues/117

When destination network is changed by the user IMO the resource selected, amount and destination address need to be reset so that they may update the UI with correct values. I have added functionality that resets the selected resource, which compels user to select the desired token again. Furthermore, the address input should be cleared if the source network is of type `EVM` however, potential destination is of a different type

Closes: #117 

## Types of changes

### DIFF
--- a/.github/workflows/cf-deploy-react.yml
+++ b/.github/workflows/cf-deploy-react.yml
@@ -3,13 +3,11 @@ on:
   push:
     branches:
       - main
-    paths:
-    - 'packages/react/**'
+      - dev
   pull_request:
     branches:
       - main
-    paths:
-      - 'packages/react/**'
+      - dev
 
 jobs:
   deploy:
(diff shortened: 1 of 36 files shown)


mpetrunic:
This should go into main branch. It's bugfix for current release^^

saadjhk:
> This should go into main branch. It's bugfix for current release.
> 
> It seems like you are going to need to create new PR and cherry pick this commit
> 
> cc @saadjhk

Sure will create separate PR

Issue #150, opened 2024-03-21
Input field jump to decimal after 1 digit input


LyonSsS:
Severity: Low | Priority: Medium

**Context**:  When using input value in the "Amount to transfer", after first digit input, the input cursor jumps to the last decimal. 
Found in PR LINK - https://github.com/sygmaprotocol/sygma-widget/pull/143 
![JumpToDecimal](https://github.com/sygmaprotocol/sygma-widget/assets/30255634/18c29765-4885-4eda-922d-6dfeaab233f8)

LyonSsS:
Tested and fixed 
![Jumping_decimal](https://github.com/sygmaprotocol/sygma-widget/assets/30255634/727fa176-fecf-49a4-80bc-479ac584dbc1)

PR #159, opened 2024-03-28
fix: easier amount input

mpetrunic:
## Related Issue Or Context

Closes: #150 

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] Documentation

## Checklist:


Lykhoyda (approved):
Looks good, please fix the tests

packages/widget/src/components/resource-amount-selector/resource-amount-selector.ts
```diff
@@ -103,6 +103,9 @@ export class ResourceAmountSelector extends BaseComponent {
   };
 
   _validateAmount(amount: string): boolean {
+    if (amount === '') {
```
wainola, line 106: if I try to pass a number like `13.45` the `.` will be treated as empty string, and the amount will be reseted to zero. Firefox treats this differently. The dot repeats the actual value that is currently int the input, and then continues. This is why this block works on firefox but not on chromium based browsers

wainola (changes requested):

packages/widget/src/components/resource-amount-selector/resource-amount-selector.ts
mpetrunic: I'm sorry, I don't follow. I just tried 13.45 on both chrome and firefox and it works perfectly

wainola: my bad, I wasn't aware that the browser was taking into consideration the localization for the type of symbol you use for decimal places, in my case `,` and not `.`, so everything is fine here. First time using `,` for decimal places though 😄

wainola:
there is a small issue with the cursor when deleting the amount, it goes to the start of the input and leave some numbers there

Lykhoyda:
@mpetrunic @wainola I think we can keep the type number for now, it's not causing big issues. If we want to change it, we can open a new issue. I will approve this one

Lykhoyda (approved):

PR #160, opened 2024-03-28
fix: Switching to a different TOKEN mapping for "From - TO" relationship is not updated in UI if amount is inserted

saadjhk:
- Added network type to address input component
- remove selected resource on destination network change

## Description

Resource selection is not updated on UI when a destination network is changed.

## Related Issue Or Context

https://github.com/sygmaprotocol/sygma-widget/issues/117

When destination network is changed by the user IMO the resource selected, amount and destination address need to be reset so that they may update the UI with correct values. I have added functionality that resets the selected resource, which compels user to select the desired token again. Furthermore, the address input should be cleared if the source network is of type `EVM` however, potential destination is of a different type

Closes: #117 

## Types of changes

### DIFF
--- a/packages/widget/src/components/transfer/fungible/fungible-token-transfer.ts
+++ b/packages/widget/src/components/transfer/fungible/fungible-token-transfer.ts
@@ -9,6 +9,7 @@ import {
   FungibleTokenTransferController,
   FungibleTransferState
 } from '../../../controllers/transfers/fungible-token-transfer';
+import '../../common/buttons/button';
 import '../../address-input';
 import '../../resource-amount-selector';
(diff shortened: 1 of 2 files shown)


wainola:
@saadjhk if I change source network, the selectors allows me to have unsoported routes, like `Cronos - Cronos`. Also if I swtich the network from the selector which in turns triggers switching in metamask, the destination selector doesn't get updated correctly. I have the impression part of the changes here undo some fixes that were merge on main. Can you confirm this?

wainola (approved):

mpetrunic (approved):

Issue #87, opened 2024-02-12
Enable adding other web3 wallets


mpetrunic:
## Implementation details
We should enable users to configure other wallets in Onboard library.

saadjhk:
@mpetrunic @MakMuftic do we have a list of wallets that we would want to enable w.r.t priorities

itsbobbyzzz168:
@saadjhk yes, list of wallets below:

EVM
1. metamask
2. Wallet Connect

Substrate
1. Polkadot.js
2. Talisman
3. Nova wallet

Overall order
1. Metamask
2. Polkadot.jjs
3. Talisman
4. Wallet connect/Nova

mpetrunic:
Hey @saadjhk the goal is just enable user to pass https://github.com/blocknative/web3-onboard/blob/develop/packages/core/src/types.ts#L25 so they can install/enable any wallet supported by the library

saadjhk:
@LyonSsS you would be able to verify this by providing your own wallet modules to `SygmaWidget` compoent

itsbobbyzzz168:
@saadjhk hi ser, just for visibility if needing QA, just easier if you add label plus move it into QA and then will see it straight away <img width="1838" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBNnRlQkE9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--e6314d6faceadada4006966f10ec1c2161a52600/Screenshot%202024-04-15%20at%2014.00.45.png" alt="Screenshot 2024-04-15 at 14.00.45.png" />

saadjhk:
@itsbobbyzzz168 got it 👍🏻

LyonSsS:
Tested on branch https://github.com/sygmaprotocol/sygma-widget  
I added a few wallets for test: 
<img width="1933" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBK3BsQkE9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--8cbe094b64ecfa1bc123e0812bfc5f0ece95de1f/image.png" alt="image.png" />
All of them work but some of the integrations from "onboard" may require extra set-up. 
[...]

PR #161, opened 2024-04-03
feat: enable user to specify wallets

saadjhk:
Add functionality to enable users to specify `onboard.js` wallet modules

## Description

Add Widget property `wallets` that accepts list of `WalletInit[]` objects from 3rd party app

## Related Issue Or Context

[Enable adding other web3 wallets](https://github.com/sygmaprotocol/sygma-widget/issues/87)

Closes: #87

## Types of changes

- [x] `WalletController` now supports ability to specify onboard wallets


packages/widget/src/controllers/wallet-manager/manager.ts
```diff
@@ -7,15 +7,40 @@ import Onboard from '@web3-onboard/core';
 import injectedModule from '@web3-onboard/injected-wallets';
 import walletConnectModule from '@web3-onboard/walletconnect';
 import type { ReactiveController, ReactiveElement } from 'lit';
+import type { WalletInit } from '@web3-onboard/common';
 
 import { utils } from 'ethers';
 import { WalletUpdateEvent, walletContext } from '../../context';
 
 export class WalletController implements ReactiveController {
+  readonly WALLET_CONNECT_PROJECT_ID = '2f5a3439ef861e2a3959d85afcd32d06';
```
wainola, line 16: should this be hardcoded?  or did you forget to delete it?

packages/widget/src/controllers/wallet-manager/manager.ts
```diff
@@ -7,15 +7,40 @@ import Onboard from '@web3-onboard/core';
...
+  /**
+   * Provides list of wallets specified by 3rd party
+   * if not, provides default list
+   * @param {{ dappUrl?: string }} options
+   * @returns {WalletInit[]}
+   */
+  getWallets(options?: { dappUrl?: string }): WalletInit[] {
+    const specifiedWallets = this.walletContext.value?.wallets;
+
+    if (!!specifiedWallets && specifiedWallets.length > 0)
```
wainola, line 31: nit: could you put at the same line the`return specifiedWallets`? if not, we could open a block.

wainola (changes requested):

wainola:
@saadjhk you are missing some changes that we currently have on main for the wallet context and the manger mainly. `connectEvmWallet` receives and object for `walletConnectOptions` as well as `appMetadata`. You can find this on main branch. In that case it would make sense to rebase from main.

packages/widget/src/controllers/wallet-manager/manager.ts
saadjhk: @wainola I have used it from [here](https://github.com/sygmaprotocol/sygma-widget/blob/13235bb6784e2d0e6249d30aa58c1beef4068358/packages/widget/src/controllers/wallet-manager/manager.ts#L101C23-L101C55)

saadjhk:
@wainola I have rebased with `dev`

packages/widget/src/controllers/wallet-manager/manager.ts
sztok7: shouldn't linter with prettier take care of this?

packages/widget/src/context/wallet.ts
```diff
@@ -57,8 +59,14 @@ export class WalletContextProvider extends BaseComponent {
   @property({ attribute: false, type: Object })
   evmWallet?: EvmWallet;
 
+  @property({ type: Array })
+  wallets?: WalletInit[];
```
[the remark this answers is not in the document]
sztok7: hey @saadjhk not sure if you still haven't pushed changes or discussed this with Marin offline (if this is the case, it's good to leave a comment, so other reviewers are aware too), but it's better that the reviewer resolves discussions https://handbook.chainsafe.io/development/development-flow/peer-reviews/author-guide#authors-can-never-ever-resolve-conversations

packages/widget/src/controllers/wallet-manager/manager.ts
```diff
@@ -7,23 +7,66 @@ import Onboard from '@web3-onboard/core';
...
+
+  /**
+   * Provides list of wallets specified by 3rd party
+   * along with default injected connector
+   * @param {{ dappUrl?: string }} options
+   * @returns {WalletInit[]}
+   */
+  getWallets(options?: {
+    walletConnectOptions?: WalletConnectOptions;
+    appMetaData?: AppMetadata;
```
mpetrunic, line 30: instead of subscribing to config context, you can just add `walletSelectorWalletConfigurations` here

packages/widget/src/controllers/wallet-manager/manager.ts
```diff
@@ -7,23 +7,66 @@ import Onboard from '@web3-onboard/core';
...
+  }): WalletInit[] {
+    // always have injected ones
+    const injected = injectedModule();
+    let wallets = [injected];
+
+    if (options?.walletConnectOptions?.projectId) {
+      // wallet connect can be directly provided by
+      // the user, do we still need to check this?
+      // seems unnecessary
```
mpetrunic, line 39: Probably not

mpetrunic (dismissed):

packages/widget/src/controllers/wallet-manager/manager.ts
```diff
@@ -18,8 +17,34 @@ export class WalletController implements ReactiveController {
...
+    let wallets = [injected];
+
+    if (
+      options?.walletSelectorWalletConfigurations &&
+      options?.walletSelectorWalletConfigurations.length > 0
+    ) {
+      wallets = wallets.concat(options.walletSelectorWalletConfigurations);
+    }
+
+    return wallets;
```
sztok7, line 42: I'd maybe rather use `const` if possible + a bit shorter `if` condition

```suggestion
    const wallets = [injected];

    if (options?.walletSelectorWalletConfigurations?.length) {
      wallets.push(...options.walletSelectorWalletConfigurations);
    }

    return wallets;
```

packages/widget/src/context/config.ts
```diff
@@ -3,14 +3,15 @@ import { createContext, ContextProvider } from '@lit/context';
...
-import type { AppMetadata } from '@web3-onboard/common';
+import type { AppMetadata, WalletInit } from '@web3-onboard/common';
 import { BaseComponent } from '../components/common/base-component';
 import type { Theme } from '../interfaces';
 
 export interface ConfigContext {
   theme?: Theme;
   walletConnectOptions?: WalletConnectOptions;
   appMetaData?: AppMetadata;
+  walletSelectorWalletConfigurations?: WalletInit[];
```
sztok7, line 14: i know you just applied @mpetrunic's suggestion, so sorry for this, but still seems a bit weird, especially having word 'wallet' appear 2 times 🤣 
how about `walletConfigModules` or just `walletModules`? i think this way it's pretty self-explanatory and a lot shorter than current one :)

sztok7 (changes requested):

packages/widget/src/context/config.ts
saadjhk: @sztok7 yeah that is a difficult name for an attribute i agree 😅

saadjhk: how about `walletInitModules` ?

mpetrunic: I'm fine with walletModules

sztok7 (approved):

wainola (approved):

PR #163, opened 2024-04-08
feat: reset fields after transfer is complete 

### LINKED ISSUE #133: Source network gets reset on starting new transfer

Lykhoyda:
closes #133

### DIFF
--- a/packages/widget/src/components/common/dropdown/dropdown.ts
+++ b/packages/widget/src/components/common/dropdown/dropdown.ts
@@ -47,17 +47,14 @@ export class Dropdown extends BaseComponent {
 
   constructor() {
     super();
-    console.log('Preselect from Constructor');
     this._setPreselectedOption();
   }
 
   _setPreselectedOption = (): void => {
     if (this.preSelectedOption) {
-      console.log('HAS PRESELECTED OPTION');
       const newOption =
         this.options.find((o) => o.name === this.preSelectedOption) || null;
 
-      console.log('Preselect from _setPreselectedOption', newOption);
       if (newOption) {
         this.selectedOption = newOption;
         this.onOptionSelected(newOption);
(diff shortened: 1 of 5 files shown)


packages/widget/src/components/common/dropdown/dropdown.ts
```diff
@@ -36,12 +36,32 @@ export class Dropdown extends BaseComponent {
...
+
   @state()
   selectedOption: DropdownOption | null = null;
 
   @property({ attribute: false })
   onOptionSelected: (option?: DropdownOption) => void = () => {};
 
+  constructor() {
+    super();
+    this._setPreselectedOption();
```
mpetrunic, line 50: isn't this suppose to be done on connected callback?

mpetrunic (approved):

packages/widget/src/components/common/dropdown/dropdown.ts
Lykhoyda: yeah, I can move it to the connected callback as well 👍

mpetrunic (approved):

packages/widget/src/controllers/transfers/fungible-token-transfer.ts
```diff
@@ -105,8 +105,10 @@ export class FungibleTokenTransferController implements ReactiveController {
     this.host.requestUpdate();
   }
 
-  reset(): void {
-    this.sourceNetwork = undefined;
+  reset({ omitSourceNetworkReset } = { omitSourceNetworkReset: false }): void {
```
wainola, line 108: if this is false by default, wouldn't make sense to define it as optional parameter and type cast the undefined? 🤔

wainola (changes requested):

wainola (commented):
Left one small comment

wainola (approved):

PR #169, opened 2024-04-12
feat: network whitelisting by User

### LINKED ISSUE #148: Enable whitelisting resources
## Implementation details

Make sure if user passed whitelistedSourceResources property that we filter available resources to only allow those. Property should probably changed to accept some identifier and not entire resource object.

Lykhoyda:
Closes #148


packages/widget/index.html
```diff
@@ -10,8 +10,15 @@
...
   </sygmaprotocol-widget>
+    <script>
+      const widget = document.getElementById('sygmaprotocolWidget');
+      widget.whitelistedSourceNetworks = ['sepolia', 'cronos'];
+      widget.whitelistedDestinationNetworks = ['mumbai', 'sepolia', 'rococo-phala'];
+      widget.whitelistedSourceResources = ['ERC20LRTest'];
+    </script>
 </body>
 
+
```
Lykhoyda, line 23: I put that for a testing purpose. 
⚠️ To be removed before the merge

packages/widget/src/context/config.ts
```diff
@@ -24,6 +27,15 @@ export class ConfigContextProvider extends BaseComponent {
...
   });
 
+  @property({ attribute: false, type: Object })
+  whitelistedSourceNetworks?: string[];
+
+  @property({ attribute: false, type: Object })
+  whitelistedDestinationNetworks?: string[];
+
+  @property({ attribute: false, type: Object })
+  whitelistedSourceResources?: string[];
```
mpetrunic, line 37: Is there a reason for putting this into context? We are passing it only to the first child (`<sygma-fungible-transfer\`), while context is used to avoid passing it trough multiple levels

mpetrunic (changes requested):
Should go onto dev branch^^

saadjhk:
> Should go onto dev branch^^

remove the `EVM` filter if merging with `dev`

Lykhoyda:
Reopen for dev branch https://github.com/sygmaprotocol/sygma-widget/pull/170

Issue #146, opened 2024-03-19
Implement network whitelisting


mpetrunic:
## Implementation details

Use widget properties whitelistedSourceNetworks and whitelistedDestinationNetworks to filter available networks. Users will be passing chainIds or network names.

PR #170, opened 2024-04-12
feat: enable developers to configure network whitelisting

### LINKED ISSUE #148: Enable whitelisting resources

Lykhoyda:
Closes #148 #146

### DIFF
--- a/packages/widget/index.html
+++ b/packages/widget/index.html
@@ -14,5 +14,4 @@
   </sygmaprotocol-widget>
 </body>
 
-
 </html>
--- a/packages/widget/index.html
+++ b/packages/widget/index.html
@@ -10,14 +10,8 @@
 </head>
 
 <body>
-  <sygmaprotocol-widget id="sygmaprotocolWidget" >
+  <sygmaprotocol-widget>
   </sygmaprotocol-widget>
-    <script>
-      const widget = document.getElementById('sygmaprotocolWidget');
-      widget.whitelistedSourceNetworks = ['sepolia', 'cronos'];
-      widget.whitelistedDestinationNetworks = ['mumbai', 'sepolia', 'cronos', 'rococo-phala'];
-      widget.whitelistedSourceResources = ['ERC20LRTest'];
-    </script>
 </body>
 
 
--- a/packages/widget/index.html
+++ b/packages/widget/index.html
@@ -10,8 +10,15 @@
 </head>
 
 <body>
-  <sygmaprotocol-widget>
+  <sygmaprotocol-widget id="sygmaprotocolWidget" >
   </sygmaprotocol-widget>
+    <script>
+      const widget = document.getElementById('sygmaprotocolWidget');
+      widget.whitelistedSourceNetworks = ['sepolia', 'cronos'];
+      widget.whitelistedDestinationNetworks = ['mumbai', 'sepolia', 'rococo-phala'];
+      widget.whitelistedSourceResources = ['ERC20LRTest'];
+    </script>
 </body>
+
 
 </html>
--- a/packages/widget/index.html
+++ b/packages/widget/index.html
@@ -15,7 +15,7 @@
     <script>
       const widget = document.getElementById('sygmaprotocolWidget');
       widget.whitelistedSourceNetworks = ['sepolia', 'cronos'];
-      widget.whitelistedDestinationNetworks = ['mumbai', 'sepolia', 'rococo-phala'];
+      widget.whitelistedDestinationNetworks = ['mumbai', 'sepolia', 'cronos', 'rococo-phala'];
       widget.whitelistedSourceResources = ['ERC20LRTest'];
     </script>
 </body>
--- a/packages/widget/src/components/transfer/fungible/fungible-token-transfer.ts
+++ b/packages/widget/src/components/transfer/fungible/fungible-token-transfer.ts
@@ -28,8 +28,6 @@ import { styles } from './styles';
 export class FungibleTokenTransfer extends BaseComponent {
   static styles = styles;
 
-  @property({ type: Array }) whitelistedSourceResources?: Array<string>;
-
   @property({ type: String })
   environment?: Environment = Environment.MAINNET;
 
--- a/packages/widget/src/components/transfer/fungible/fungible-token-transfer.ts
+++ b/packages/widget/src/components/transfer/fungible/fungible-token-transfer.ts
@@ -48,12 +48,11 @@ export class FungibleTokenTransfer extends BaseComponent {
 
   connectedCallback(): void {
     super.connectedCallback();
-    void this.transferController.init(
-      this.environment!,
-      this.whitelistedSourceNetworks,
-      this.whitelistedDestinationNetworks,
-      this.whitelistedSourceResources
-    );
+    void this.transferController.init(this.environment!, {
+      whitelistedSourceNetworks: this.whitelistedSourceNetworks,
+      whitelistedDestinationNetworks: this.whitelistedDestinationNetworks,
+      whitelistedSourceResources: this.whitelistedSourceResources
+    });
   }
 
   private onClick = (): void => {
--- a/packages/widget/src/components/transfer/fungible/fungible-token-transfer.ts
+++ b/packages/widget/src/components/transfer/fungible/fungible-token-transfer.ts
@@ -31,13 +31,13 @@ export class FungibleTokenTransfer extends BaseComponent {
   @property({ type: String })
   environment?: Environment = Environment.MAINNET;
 
-  @property({ attribute: false, type: Object })
+  @property({ type: Object })
   whitelistedSourceNetworks?: string[];
 
-  @property({ attribute: false, type: Object })
+  @property({ type: Object })
   whitelistedDestinationNetworks?: string[];
 
-  @property({ attribute: false, type: Object })
+  @property({ type: Object })
   whitelistedSourceResources?: string[];
 
   @property({ type: Object })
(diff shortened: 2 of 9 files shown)


Lykhoyda:
Is there a reason for putting this into context? We are passing it only to the first child (<sygma-fungible-transfer\), while context is used to avoid passing it trough multiple levels

@mpetrunic I think the main reason, for now, is consistency as we have other config params provided by the user in "sygma-config-context-provider", so to keep the centralized place for all the user params. I also can imagine that we could need this information in other components in the future, so probably makes sense to have it all together, although indeed it will be easier to pass the props.

Lykhoyda:
To test this story make changes in index.html file by adding or removing the networks or resources:
```html
    <script>
      const widget = document.getElementById('sygmaprotocolWidget');
      widget.whitelistedSourceNetworks = ['sepolia', 'cronos'];
      widget.whitelistedDestinationNetworks = ['mumbai', 'sepolia', 'rococo-phala'];
      widget.whitelistedSourceResources = ['ERC20LRTest'];
    </script>
    ```

**Comment by mpetrunic** [comment]
> Is there a reason for putting this into context? We are passing it only to the first child (<sygma-fungible-transfer), while context is used to avoid passing it trough multiple levels
> 
> @mpetrunic I think the main reason, for now, is consistency as we have other config params provided by the user in "sygma-config-context-provider", so to keep the centralized place for all the user params. I also can imagine that we could need this information in other components in the future, so probably makes sense to have it all together, although indeed it will be easier to pass the props.

I would keep it out of the context until there is need for it. Makes testing easier and reduces chance of additional rerenders (context is event based)

File: packages/widget/index.html [hunk 1]
```diff
@@ -10,8 +10,15 @@
...
 <body>
-  <sygmaprotocol-widget>
+  <sygmaprotocol-widget id="sygmaprotocolWidget" >
   </sygmaprotocol-widget>
+    <script>
+      const widget = document.getElementById('sygmaprotocolWidget');
+      widget.whitelistedSourceNetworks = ['sepolia', 'cronos'];
+      widget.whitelistedDestinationNetworks = ['mumbai', 'sepolia', 'cronos', 'rococo-phala'];
+      widget.whitelistedSourceResources = ['ERC20LRTest'];
+    </script>
```
mpetrunic, line 20: do we wanna keep this?

packages/widget/src/context/config.ts
```diff
@@ -1,17 +1,20 @@
...
+import { BaseComponent } from '../components/common/base-component';
 
 export interface ConfigContext {
   theme?: Theme;
   walletConnectOptions?: WalletConnectOptions;
   appMetaData?: AppMetadata;
   walletModules?: WalletInit[];
+  whitelistedSourceNetworks?: string[];
+  whitelistedDestinationNetworks?: string[];
+  whitelistedSourceResources?: string[];
```
mpetrunic, line 17: don't think you need that anymore?

packages/widget/src/controllers/transfers/fungible-token-transfer.ts
```diff
@@ -183,13 +188,46 @@ export class FungibleTokenTransferController implements ReactiveController {
...
+    return whitelistedNetworks.some(
+      (networkName) => networkName.toLowerCase() === network.name.toLowerCase()
+    );
+  };
+
+  async init(
+    env: Environment,
+    whitelistedSourceNetworks?: string[],
+    whitelistedDestinationNetworks?: string[],
+    whitelistedSourceResources?: string[]
```
mpetrunic, line 212: It would be more ergonomic for us to move this into object `options: {whitelistedSourceNmetworks, whitelistedDestinationNetworks...}` as most IDEs would show you name of the properties

packages/widget/tests/unit/components/transfer/fungible/fungible-token-transfer.test.ts
```diff
@@ -101,4 +112,71 @@ describe('Fungible token Transfer', function () {
...
+        evmWallet: {
+          address: connectedAddress,
+          providerChainId: 11155111,
+          provider: getMockedEvmWallet().provider
+        }
+      })
+    );
+
+    // Set Source and Destination Networks
+    fungibleTransfer.transferController.onSourceNetworkSelected(cronosNetwork);
```
mpetrunic, line 146: Ain't you suppose to check before this that `fungibleTransfer.transferController.supportedSource(Destination)Networks˙is correctly whitelisted?

packages/widget/src/controllers/transfers/fungible-token-transfer.ts
```diff
@@ -167,8 +172,8 @@ export class FungibleTokenTransferController implements ReactiveController {
    * Infinite Try/catch wrapper around
    * {@link Config} from `@buildwithsygma/sygma-sdk-core`
    * and emits a {@link SdkInitializedEvent}
-   * @param {number} time to wait before retrying request in ms
    * @returns {void}
+   * @param retryMs
```
mpetrunic, line 176: maybe keep the original comment 😄

packages/widget/tests/unit/components/transfer/fungible/fungible-token-transfer.test.ts
```diff
@@ -101,4 +112,71 @@ describe('Fungible token Transfer', function () {
...
+
+    const [sygmaSourceNetwork, sygmaDestinationNetwork] =
+      fungibleTransfer.shadowRoot!.querySelectorAll('sygma-network-selector');
+
+    const resourceSelector = fungibleTransfer.shadowRoot!.querySelector(
+      'sygma-resource-amount-selector'
+    ) as ResourceAmountSelector;
+
+    // Wait for sdk init
+    await aTimeout(1000);
```
mpetrunic, line 161: this is not acceptable. Better option would be to mock SDK (init method). and wait for field to update

mpetrunic (changes requested):

packages/widget/index.html
Lykhoyda: it's just for testing the PR, I will remove before merge

packages/widget/src/controllers/transfers/fungible-token-transfer.ts
Lykhoyda: good idea 👍

packages/widget/tests/unit/components/transfer/fungible/fungible-token-transfer.test.ts
mpetrunic: FWIW, I've discussed with @saadjhk and he will open issue to write tests for fungible token transfer with mocks

Lykhoyda: @mpetrunic I added the missing check from your previous comment. Leaving timeout for now

mpetrunic: Added fix to remove timeouts. Timeouts should never be merged!

mpetrunic: Added fix to remove timeouts as timeouts should never be merged!

Lykhoyda: @mpetrunic thank you so much for the help with mocking 🙏🏻🙏🏻🙏🏻

packages/widget/src/components/transfer/fungible/fungible-token-transfer.ts
```diff
@@ -25,11 +25,18 @@ import { styles } from './styles';
 export class FungibleTokenTransfer extends BaseComponent {
   static styles = styles;
 
-  @property({ type: Array }) whitelistedSourceResources?: Array<string>;
-
   @property({ type: String })
   environment?: Environment = Environment.MAINNET;
 
+  @property({ attribute: false, type: Object })
```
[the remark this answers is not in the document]
Lykhoyda: I'm not sure, about the behavior in react. My understanding is that we are using the attribute true when the value might be used in HTML for example as a value or disabled or we will pass it to the ref. But if we map through the list of whitelisted networks then it should be used as attribute: false. Correct me if I'm wrong

LyonSsS:
Tested on PR with a mix of valid and invalid or missing form sharedconfig.json data
Valid
image
Valid - since Mumbai was removed and Sepolia can't send to Sepolia 
image
image
image
Valid 
image
image

packages/widget/tests/unit/components/transfer/fungible/fungible-token-transfer.test.ts
mpetrunic: All good, don't hesitate to ask for help in channel when you encounter stuff like that :)

Lykhoyda:
> In react if you update, for example, whitelistedSourceNetworks. That update won't be reflected in the component. Fungible token transfer component should listen for field updates and reinit transfer controller on change

Got it, removed the attribute: false 👍

mpetrunic:
> > In react if you update, for example, whitelistedSourceNetworks. That update won't be reflected in the component. Fungible token transfer component should listen for field updates and reinit transfer controller on change
> 
> Got it, removed the attribute: false 👍

@Lykhoyda Please reread my comment haha

> Fungible token transfer component should listen for field updates and reinit transfer controller on change

packages/widget/index.html
mpetrunic, line 1: Please revert this file so we can approve

wainola (approved):

mpetrunic (approved):

Issue #157, opened 2024-03-27
Display gas fee estimation


mpetrunic:
depends on #153 

## Implementation details

In addition to fee, we should display estimation on how much gas (in native currency) + dollar value should transfer cost

MakMuftic:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Sygma-Protocol-62f126a54b7d5a55203732f0/issues/sygmaprotocol/sygma-widget/157?planning-poker) with Zenhub @saadjhk @wainola @mpetrunic @Lykhoyda

saadjhk:
@mpetrunic @MakMuftic I'd like to know what service provider are we going to use in order to show value in fiat

mpetrunic:
value in fiat is blocked by #153

saadjhk:
@mpetrunic I went through the sdk and I see that there isn't any method that provides estimation of the contract call, I will believe we should add it there first?

mpetrunic:
> @mpetrunic I went through the sdk and I see that there isn't any method that provides estimation of the contract call, I will believe we should add it there first?

But in the widget you have transaction(+approvals) that will be executed, just can use ethers to estimate.

saadjhk:
@mpetrunic so the flow should show estimation for each transaction (including all approvals)?

mpetrunic:
> @mpetrunic so the flow should show estimation for each transaction (including all approvals)?

Well approvals are also cost of transfer 😅

LyonSsS:
<img width="564" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBeVBFQkE9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--d0d771675c82a0b3ba7edf3ca5ddd113d6631f16/image.png" alt="image.png" />
All without the dollar value ( which is currently blocked)

PR #174, opened 2024-04-16
feat: display transfer gas cost estimation

saadjhk:
## Description
Display gas estimation of transactions that will be executed

## Related Issue Or Context
Closes: #157

## How Has This Been Tested? Testing details.
- [x] Mocked gas fee test 

## Types of changes
- [x] Transactions gas fee estimations for evm and substrate networks

## Checklist:
- [x] Add tests.

### DIFF
--- a/packages/widget/src/components/transfer/fungible/transfer-detail/styles.ts
+++ b/packages/widget/src/components/transfer/fungible/transfer-detail/styles.ts
@@ -1,6 +1,12 @@
 import { css } from 'lit';
 
 export const styles = css`
+  .transferDetail {
+    gap: var(--spacing-md, 8px);
+    display: flex;
+    flex-direction: column;
+  }
+
   .transferDetailContainer {
     display: flex;
     font-size: 0.75rem;
--- a/packages/widget/src/components/transfer/fungible/transfer-detail/styles.ts
+++ b/packages/widget/src/components/transfer/fungible/transfer-detail/styles.ts
@@ -2,7 +2,7 @@ import { css } from 'lit';
 
 export const styles = css`
   .transferDetail {
-    gap: var(--spacing-md, 8px);
+    gap: 0.5rem;
     display: flex;
     flex-direction: column;
   }
--- a/packages/widget/src/components/transfer/fungible/transfer-detail/transfer-detail.ts
+++ b/packages/widget/src/components/transfer/fungible/transfer-detail/transfer-detail.ts
@@ -10,6 +10,7 @@ import type { HTMLTemplateResult } from 'lit';
 import { html } from 'lit';
 import { when } from 'lit/directives/when.js';
 import { customElement, property } from 'lit/decorators.js';
+import { BigNumber } from 'ethers';
 import { tokenBalanceToNumber } from '../../../../utils/token';
 import { BaseComponent } from '../../../common/base-component';
 import { styles } from './styles';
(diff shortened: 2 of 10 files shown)


packages/widget/src/components/transfer/fungible/transfer-detail/transfer-detail.ts
```diff
@@ -49,19 +56,46 @@ export class FungibleTransferDetail extends BaseComponent {
...
+    const { symbol, decimals } = this.getGasFeeParams();
+
+    let gasFee;
+    if (decimals && this.estimatedGasFee) {
+      gasFee = tokenBalanceToNumber(this.estimatedGasFee, decimals, 4);
+
+      return `${gasFee} ${symbol}`;
+    }
+
+    return '';
```
mpetrunic, line 96:
```suggestion
    return 'calculating...';
```
?

mpetrunic (approved):

packages/widget/src/components/transfer/fungible/transfer-detail/styles.ts
```diff
@@ -1,6 +1,12 @@
 import { css } from 'lit';
 
 export const styles = css`
+  .transferDetail {
+    gap: var(--spacing-md, 8px);
```
wainola, line 5: do we have this var defined? Also pls change px to rem

packages/widget/src/components/transfer/fungible/transfer-detail/transfer-detail.ts
```diff
@@ -49,19 +56,46 @@ export class FungibleTransferDetail extends BaseComponent {
     }
   }
 
-  getFee(): string {
+  getGasFeeParams(): {
+    decimals?: number;
+    symbol: string;
+  } {
+    let decimals = undefined;
```
wainola, line 63: nit and question: if we use default values for the parameters, would this two declarations make sense? 🤔

packages/widget/src/components/transfer/fungible/transfer-detail/transfer-detail.ts
```diff
@@ -49,19 +56,46 @@ export class FungibleTransferDetail extends BaseComponent {
...
     }
 
     return `${_fee} ${symbol}`;
   }
 
+  getEstimatedGasFee(): string {
+    if (!this.estimatedGasFee) return '';
+    const { symbol, decimals } = this.getGasFeeParams();
+
+    let gasFee;
```
wainola, line 89: I would expect here the linter to suggest to use const, since this variable is not being used outside the block. I think you can just declare `const gasFee` in the block, and then return it.

wainola (changes requested):

packages/widget/src/components/transfer/fungible/transfer-detail/transfer-detail.ts
saadjhk: what do you mean by default values?

wainola: resolved, nvm this comment

packages/widget/src/controllers/transfers/fungible-token-transfer.ts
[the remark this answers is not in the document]
mpetrunic: Will be resolved in separe PR. Related issue #177

mpetrunic (approved):

wainola (dismissed):
just a small nit: `Bridge fee` appears to be render before `Amount to receive` and `Gas fee`

saadjhk:
> just a small nit: `Bridge fee` appears to be render before `Amount to receive` and `Gas fee`

Fixed

wainola (approved):

mpetrunic (approved):

PR #176, opened 2024-04-17
fix: bug where user could proceed with invalid destination address

### LINKED ISSUE #164: Address validation, still allows transfer to move forward if address format is incorrect
Priority: Medium | Severity: Medium

## Short context:
"Send to" field validation should block transfer if an address is invalid. 
Currently it is not.

## STR:
1. Connect an EVM wallet and select source Sepolia.
2. Select destination as Cronos and select the available token type from the drop down.
3. Input a valid amount.
4. Input an invalid ETH format address into the "Send to" field ( use a substrate format one)

Exp: The Allow spending/Transfer button should be blocked ( disable).
Act: The user can use the Allow Spending & Transfer button to move forward with the flow. 

Note: This works the other way around - entering an EVM address for the Substrate format. 
image
image

Lykhoyda:
closes #164

### DIFF
--- a/packages/widget/src/components/address-input/address-input.ts
+++ b/packages/widget/src/components/address-input/address-input.ts
@@ -23,7 +23,7 @@ export class AddressInput extends BaseComponent {
   onAddressChange: (address: string, errorMessage?: string | null) => void =
     () => {};
 
-  @property()
+  @property({ attribute: false })
   networkType: Network = Network.EVM;
 
   @state()
--- a/packages/widget/src/components/address-input/address-input.ts
+++ b/packages/widget/src/components/address-input/address-input.ts
@@ -63,7 +63,7 @@ export class AddressInput extends BaseComponent {
           <span>Send to </span>
           ${when(
             this.errorMessage,
-            () => html` <span class="errorMessage">${this.errorMessage}</span>`
+            () => html`<span class="errorMessage">${this.errorMessage}</span>`
           )}</label
         >
         <textarea
--- a/packages/widget/src/components/address-input/address-input.ts
+++ b/packages/widget/src/components/address-input/address-input.ts
@@ -6,7 +6,7 @@ import { when } from 'lit/directives/when.js';
 
 import type { PropertyValues } from '@lit/reactive-element';
 import { validateAddress } from '../../utils';
-import { BaseComponent } from '../common/base-component/base-component';
+import { BaseComponent } from '../common';
 
 import { styles } from './styles';
 
@@ -20,11 +20,10 @@ export class AddressInput extends BaseComponent {
   address: string = '';
 
   @property({ attribute: false })
-  onAddressChange: (address: string) => void = () => {};
+  onAddressChange: (address: string, errorMessage?: string | null) => void =
+    () => {};
 
-  @property({
-    type: String
-  })
+  @property({ type: String })
   networkType: Network = Network.EVM;
 
   @state()
@@ -49,7 +48,7 @@ export class AddressInput extends BaseComponent {
 
     this.errorMessage = validateAddress(trimedValue, this.networkType);
 
-    this.onAddressChange(trimedValue);
+    this.onAddressChange(trimedValue, this.errorMessage);
   };
 
   protected updated(changedProperties: PropertyValues): void {
@@ -60,7 +59,7 @@ export class AddressInput extends BaseComponent {
   }
 
   render(): HTMLTemplateResult {
-    return html` <section class="inputAddressSection">
+    return html`<section class="inputAddressSection">
       <div class="inputAddressContainer">
         <label class="labelContainer">
           <span>Send to </span>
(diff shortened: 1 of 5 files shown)


packages/widget/src/components/transfer/fungible/transfer-button/transfer-button.ts
```diff
@@ -7,13 +7,17 @@ import type { Button } from '../../../common';
...
 
 const enabledStates = [
-  FungibleTransferState.WRONG_CHAIN,
-  FungibleTransferState.WALLET_NOT_CONNECTED,
   FungibleTransferState.PENDING_APPROVALS,
   FungibleTransferState.PENDING_TRANSFER,
   FungibleTransferState.COMPLETED
 ];
 
+const disabledState = [
```
mpetrunic, line 15: wrong_chain and WALLET_NOT_CONNECTED shouldn't be here as in those cases click on the button will prompt wallet connection/chain switch.

I don't think you even need this var as any state that's not in enabled should be disabled state

packages/widget/src/components/transfer/fungible/transfer-button/transfer-button.ts
```diff
@@ -28,9 +32,12 @@ export class FungibleTransferButton extends BaseComponent {
   @property({ type: Object })
   onClick: () => void = () => {};
 
+  handleDisabledState = (): boolean =>
+    !enabledStates.includes(this.state) || disabledState.includes(this.state);
+
   render(): HTMLTemplateResult {
     return html`<sygma-action-button
-      .disabled=${!enabledStates.includes(this.state)}
+      .disabled=${this.handleDisabledState()}
```
mpetrunic, line 40: revert to previous logic, new one isn't making sense haha

mpetrunic (changes requested):

packages/widget/src/components/transfer/fungible/transfer-button/transfer-button.ts
Lykhoyda: Now I got the logic, that we care only about "enabled" statuses, and all others are treated as errors 👍

packages/widget/src/components/address-input/address-input.ts
```diff
@@ -20,11 +20,10 @@ export class AddressInput extends BaseComponent {
...
 
   @property({ attribute: false })
-  onAddressChange: (address: string) => void = () => {};
+  onAddressChange: (address: string, errorMessage?: string | null) => void =
+    () => {};
 
-  @property({
-    type: String
-  })
+  @property()
```
wainola, line 26:
```suggestion
  @property({ type: Object })
```

wainola (changes requested):

packages/widget/src/components/address-input/address-input.ts
Lykhoyda: I think it's not necessary as we are not using it as HTML attribute. I added { attribute: false } to indicate that we use it in code internally outside the render

wainola (approved):

packages/widget/src/components/transfer/fungible/transfer-button/transfer-button.ts
```diff
@@ -54,6 +54,10 @@ export class FungibleTransferButton extends BaseComponent {
           ],
           [FungibleTransferState.WALLET_NOT_CONNECTED, () => 'Connect Wallet'],
           [FungibleTransferState.WRONG_CHAIN, () => 'Switch chain'],
+          [
+            FungibleTransferState.WRONG_DESTINATION_ADDRESS,
+            () => 'Wrong Address'
```
mpetrunic, line 59:
```suggestion
            () => 'Invalid Address'
```

packages/widget/src/controllers/transfers/fungible-token-transfer.ts
```diff
@@ -35,6 +36,7 @@ export class FungibleTokenTransferController implements ReactiveController {
   public waitingTxExecution: boolean = false;
   public transferTransactionId?: string;
   public errorMessage: string | null = null;
+  public invalidDestinationAddressErrorMessage: string | null = null;
```
mpetrunic, line 39: since controller doesn't really care about address. Maybe we could just have null state for desinationAddress:
```
empty string = missing address
null/undefined = invalid address
anything else valid address
```
wdyt?

Lykhoyda: @mpetrunic In general, I like the idea as I also would like to avoid sending the state back to the parent. The only issue I see is that if I enter the substrate address when the EVM will be expected, I will have an error on the field level, but the button will have an "enabled" state, which is incorrect. So I will call "validateAddress" one more time. I pushed the code with the changes

LyonSsS (approved):
Tested on preview link  from PR branch and passed.

Issue #190, opened 2024-05-08
Gas Fee estimation behavior, is not consistent for the actual transfer TX


LyonSsS:
**Context**: When a user transfers a token from one Chain to another, there is a 2 step process, where he signs for allow spending and then for the actual token transfer tx. The widget does not apply the same logic for the Gass Fee estimation behavior when we reach the transfer TX. 

![image](https://github.com/sygmaprotocol/sygma-widget/assets/30255634/633e8463-55a8-45d5-aacb-c960bdf83e5e)

MakMuftic:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Sygma-Protocol-62f126a54b7d5a55203732f0/issues/sygmaprotocol/sygma-widget/190?planning-poker) with Zenhub @saadjhk @wainola

PR #204, opened 2024-05-29
fix: Gas Fee estimation behavior, is not consistent for the actual transfer TX

saadjhk:
## Description

Gas estimation for transfer transaction was not calculated as the estimate gas function was called when the transfer state returned from `getTransferState` was the one which had no case handled in `estimateEvmGas`

## Related Issue Or Context

#190 

Closes: #190 

## Types of changes

### DIFF
--- a/packages/widget/src/controllers/transfers/evm/execute.ts
+++ b/packages/widget/src/controllers/transfers/evm/execute.ts
@@ -28,7 +28,6 @@ export async function executeNextEvmTransaction(
       this.host.requestUpdate();
       await tx.wait();
       this.pendingEvmApprovalTransactions.shift();
-      await this.estimateGas();
     } catch (e) {
       console.log(e);
       this.errorMessage = 'Approval transaction reverted or rejected';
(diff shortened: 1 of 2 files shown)


packages/widget/src/controllers/transfers/evm/execute.ts
```diff
@@ -28,14 +28,14 @@ export async function executeNextEvmTransaction(
...
       this.pendingEvmApprovalTransactions.shift();
-      await this.estimateGas();
     } catch (e) {
       console.log(e);
       this.errorMessage = 'Approval transaction reverted or rejected';
     } finally {
       this.waitingUserConfirmation = false;
       this.waitingTxExecution = false;
       this.host.requestUpdate();
+      this.estimateGas();
```
wainola, line 38:
```suggestion
      await this.estimateGas();
```

wainola (changes requested):

wainola (approved):

Lykhoyda (approved):

